PR #1, opened 2023-06-30
New Rule Proj0001: .NET project file could not be located

Corniel:
So that other rules don't have to crash (or all report) that the .NET project file could not be resolved, this rule handles that scenario for them.

### DIFF
--- a/rules/Proj0001.md
+++ b/rules/Proj0001.md
@@ -1,0 +1,7 @@
+# Proj0001: .NET project file could not be located
+Roslyn officially does not support the analysis on .NET project files. Therefor
+all rules in this package try to resolve the project file on disk based on
+(naming and location) conventions.
+
+This rule reports when the project file could not be located, due to whatever
+reason.
--- a/rules/Proj0001.md
+++ b/rules/Proj0001.md
@@ -1,6 +1,6 @@
 # Proj0001: .NET project file could not be located
-Roslyn officially does not support the analysis on .NET project files. Therefor
-all rules in this package try to resolve the project file on disk based on
+Roslyn officially does not support the analysis on .NET project files.
+Therefore all rules in this package try to resolve the project file based on
 (naming and location) conventions.
 
 This rule reports when the project file could not be located, due to whatever
--- a/specs/DotNetProjectFile.Analyzers.Specs/Extensions/Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Extensions/Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer.cs
@@ -1,0 +1,26 @@
+﻿using CodeAnalysis.TestTools;
+using CodeAnalysis.TestTools.Contexts;
+
+namespace Microsoft.CodeAnalysis.Diagnostics;
+
+internal static class DiagnosticAnalyzerExtensions
+{
+    public static ProjectAnalyzerVerifyContext ForProject(this DiagnosticAnalyzer analyzer, string fileName)
+    {
+        var name = Path.GetFileNameWithoutExtension(fileName);
+        var file = new FileInfo($"../../../../../projects/{name}/{fileName}proj");
+        var context = analyzer.ForProject(file);
+
+        return context with
+        {
+            IgnoredDiagnostics = DiagnosticIds.Empty.AddRange(
+                "BC50001", // Unused import statement.
+
+                "CS8019", // Unnecessary using directive.
+                "CS8933", // The using directive appeared previously as global using
+                
+                "??????"
+            ),
+        };
+    }
+}
--- a/specs/DotNetProjectFile.Analyzers.Specs/FluentAssertions/FindsExtensions.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/FluentAssertions/FindsExtensions.cs
@@ -6,7 +6,7 @@ namespace FluentAssertions;
 internal static class FindsExtensions
 {
     public static void HasNoIssues(this ProjectAnalyzerVerifyContext context)
-        => context.HasIssues();
+        => context.HasIssues(Array.Empty<Issue>());
 
     public static void HasIssue(this ProjectAnalyzerVerifyContext context, Issue issue)
         => context.HasIssues(issue);
--- a/specs/DotNetProjectFile.Analyzers.Specs/FluentAssertions/FindsExtensions.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/FluentAssertions/FindsExtensions.cs
@@ -1,0 +1,23 @@
+﻿using CodeAnalysis.TestTools.Contexts;
+using Microsoft.CodeAnalysis;
+
+namespace FluentAssertions;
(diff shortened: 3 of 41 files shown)


src/DotNetProjectFile.Analyzers/Xml/PackageReference.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public sealed class PackageReference : Node
+{
+    public PackageReference(XElement element) : base(element)
+    {
+    }
+
+    public string? Include => GetAttribute();
+
+    public string? Version => GetAttribute();
```
CptWesley, line 13: Is it intentional that `Include` and `Version` are the same?

Corniel:
``` C#
  public string? GetAttribute([CallerMemberName] string? propertyName = null)
        => Element.Attribute(propertyName)?.Value;
 ```
 
So the return the value of different attributes.

Comment (CptWesley) [reply src/DotNetProjectFile.Analyzers/Xml/PackageReference.cs:13]: In that case I think the properties shouldn't be nullable.

File: rules/Proj0001.md [hunk 2]
```diff
@@ -0,0 +1,7 @@
+# Proj0001: .NET project file could not be located
+Roslyn officially does not support the analysis on .NET project files. Therefor
```
LauraXiulan, line 2: Therefore

rules/Proj0001.md
```diff
@@ -0,0 +1,7 @@
...
+all rules in this package try to resolve the project file on disk based on
```
LauraXiulan, line 3: This sentence reads a bit weird with the two times 'on'.

src/DotNetProjectFile.Analyzers/tools/uninstall.ps1
```diff
@@ -0,0 +1,10 @@
+param($installPath, $toolsPath, $package, $project)
+
+$analyzersPath = Split-Path -Path $toolsPath -Parent
+$analyzersPath = Join-Path $toolsPath "analyzers"
+$analyzerFilePath = Join-Path $analyzersPath "DotNetProjectFile.Analyzers.dll"
+try {
```
LauraXiulan, line 6: Why the try catch if you don't do anything with the catch?

src/DotNetProjectFile.Analyzers/Xml/Project.cs
```diff
@@ -0,0 +1,54 @@
...
+        if (compilation.SourceModule.ContainingAssembly is { } assembly)
+        {
+            var directories = assembly.Locations.Select(l => l.SourceTree?.FilePath)
+                .Where(p => p is { })
+                .Select(p => new FileInfo(p).Directory)
+                .Distinct(DirectoryEqualityComparer.Instance)
+                .Where(f => f.Exists)
+                .ToArray();
+
+            var project = directories.Select(d => GetFile(d, $"{assembly.Name}.??proj"))
```
LauraXiulan, line 29: What does this ??proj do?

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -0,0 +1,37 @@
+﻿namespace DotNetProjectFile;
+
+public static class Rule
+{
+    public static DiagnosticDescriptor ProjectFileCouldNotBeLocated => New(
+        id: 0001,
+        title: ".NET project file could not be located",
+        message: "The project file '{0}' could not be located.",
+        description: "In order to make this rules work, the project file should be located.",
```
LauraXiulan, line 9: these rules

src/DotNetProjectFile.Analyzers/Diagnostics/Category.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace DotNetProjectFile.Diagnostics;
+
+public enum Category
+{
+    Clairity,
```
LauraXiulan, line 5: Clarity

specs/DotNetProjectFile.Analyzers.Specs/FluentAssertions/FindsExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using CodeAnalysis.TestTools.Contexts;
+using Microsoft.CodeAnalysis;
+
+namespace FluentAssertions;
+
+internal static class FindsExtensions
+{
+    public static void HasNoIssues(this ProjectAnalyzerVerifyContext context)
+        => context.HasIssues();
```
LauraXiulan, line 9: This is a bit confusing, it has no issues when it has issues without parameter?

specs/DotNetProjectFile.Analyzers.Specs/Extensions/Microsoft.CodeAnalysis.Diagnostics.DiagnosticAnalyzer.cs
```diff
@@ -0,0 +1,26 @@
...
+using CodeAnalysis.TestTools.Contexts;
+
+namespace Microsoft.CodeAnalysis.Diagnostics;
+
+internal static class DiagnosticAnalyzerExtensions
+{
+    public static ProjectAnalyzerVerifyContext ForProject(this DiagnosticAnalyzer analyzer, string fileName)
+    {
+        var name = Path.GetFileNameWithoutExtension(fileName);
+        var file = new FileInfo($"../../../../../projects/{name}/{fileName}proj");
```
LauraXiulan, line 11: Does this mean that the fileName always has the .cs/.vb extension? What happens if it is called with another extension?

src/DotNetProjectFile.Analyzers/tools/uninstall.ps1
Corniel: This is what Microsoft suggests. So I just followed that:

https://learn.microsoft.com/en-us/nuget/guides/analyzers-conventions#install-and-uninstall-scripts

rules/Proj0001.md
Corniel: I dropped "on disk"

src/DotNetProjectFile.Analyzers/Xml/Project.cs
Corniel: Matching both .csproj (C#) as vbproj (VB.NET). It is a bit lazy, so I could be improved on.

specs/DotNetProjectFile.Analyzers.Specs/FluentAssertions/FindsExtensions.cs
Corniel: `HasIssues` has `param Issue[] issues`, so without params is equivalent to no issues.

Corniel: I've tried to make that easier to grasp.

src/DotNetProjectFile.Analyzers/Xml/PackageReference.cs
Corniel: What do you mean? Null means: attribute or element does not exist.

PR #34, opened 2023-07-10
FP Proj0007: `<Import />` should not be considered an empty node

Corniel:
The  `<Import />` never has child nodes (only the `Project` attribute). Long story short: the node never be considered an empty node.


CptWesley:
Isn't this also the case for a few other tags? Or is this only a problem for `Import`?

Corniel:
> Isn't this also the case for a few other tags? Or is this only a problem for `Import`?

As far as I know, it is.

PR #42, opened 2023-07-11
New Rule Proj2001: Define data in a resource file

Corniel:
# Proj2001: Define data in a resource file
A resource file without `<data>` elements is of no use.

## Non-compliant
``` XML
<root>
 
</root>
```

## Compliant
``` XML
<root>
  <!-- ... -->
  <data name="King" xml:space="preserve">
    <value>Koning</value>
  </data>
  <data name="Queen" xml:space="preserve">
    <value>Koningin</value>
  </data>
</root>
```


CptWesley (approved):
LGTM

Issue #15, opened 2023-07-08
New rule: Don't specify properties twice


Corniel:
To prevent mistakes, properties (members of `<PropertyGroup>`) should only be defined once. Also, it is without meaning to specify both `<TargetFramework>` and `<TargetFrameworks>`.

PR #45, opened 2023-07-11
New Rule Proj0010: Define the OutputType explicitly

CptWesley:
Not defining the `OutputType` property leads to the default output type `Library` which is a rather opaque, yet very impactful default setting.


projects/CompliantVB/CompliantVB.vbproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <RootNamespace>CompliantVB</RootNamespace>
     <TargetFramework>netstandard2.0</TargetFramework>
+	<OutputType>Library</OutputType>
```
Corniel, line 6: indenting is off.

rules/Proj0010.md
```diff
@@ -0,0 +1,26 @@
+# Proj0010: Define the OutputType explicitly
```
Corniel, line 1: output type are two words. (it is the term referred to, not the `<OutputType>` node.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/DefineOutputType.cs
```diff
@@ -0,0 +1,23 @@
...
+public sealed class DefineOutputType : MsBuildProjectFileAnalyzer
+{
+    public DefineOutputType() : base(Rule.DefineOutputType) { }
+
+    protected override void Register(ProjectFileAnalysisContext context)
+    {
+        if (context.Project.IsProject)
+        {
+            var outputTypes = context.Project
+                .AncestorsAndSelf()
```
Corniel, line 13: I would argue that it than also should not be a an inherited setting.

CptWesley: I would argue that it should be. I personally like to simply create different props files for different purposes (e.g. 1 for packages and 1 for tests) that each specify the `OutputType` for their type. Of course enforcing that there is only 1 definition would be very useful, but that is related to #15 .

Corniel: Also a fair point. But in that case for a property like this (and also for `<TargetFramework>` or `<IsTestProject>`) you could/should argue that those properties should then not be overridden: if defined in a prop, leave it as it. I'm not sure yet if that should be part of this rule (for `<RargetOutput>`) or not.

projects/PackagesWithAnalyzers/PackagesWithAnalyzers.csproj
```diff
@@ -37,7 +37,7 @@
     <PackageReference Include="Ardalis.ApiEndpoints.CodeAnalyzers" Version="*" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
     <PackageReference Include="FakeItEasy.Analyzer.CSharp" Version="*" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
     <PackageReference Include="Lucene.Net.Analysis.Common" Version="4.8.0-beta00016" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
-    <PackageReference Include="Libplanet.Analyzers" Version="*" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
+    <PackageReference Include="Libplanet.Analyzers" Version="2.5.0" PrivateAssets="All" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
```
Corniel, line 40: Why is this fixed?

CptWesley: `Libplanet.Analyzers` has a dependency on `Libplanet`, meaning if `*` pulls in a newer version than the `Libplanet` version, the build task will start complaining that the version of `Libplanet` was downgraded.

PR #46, opened 2023-07-12
New Rule Proj2000: Embed valid resource files

Corniel:
# Proj2000: Embed valid resources
A resource file should be a valid RESX file.

## Compliant
``` XML
<root>
  <resheader name="resmimetype">
    <value>text/microsoft-resx</value>
  </resheader>
  <resheader name="reader">
    <value>System.Resources.ResXResourceReader, System.Windows.Forms</value>
  </resheader>
  <resheader name="writer">
    <value>System.Resources.ResXResourceWriter, System.Windows.Forms</value>
  </resheader>
  <data name="King" xml:space="preserve">
    <value>Koning</value>
  </data>
  <data name="Queen" xml:space="preserve">
    <value>Koningin</value>
  </data>
</root>
```


rules/Proj2000.md
```diff
@@ -0,0 +1,27 @@
+# Proj2000: Embed valid resources
+A resource file should be a valid RESX file.
+
+## Compliant
+``` XML
+<?xml version="1.0" encoding="utf-8"?>
```
LauraXiulan, line 6: How much data should be in the RESX file to be regarded as compliant? Would just a resheader be enough or does it have to be specific values?

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -137,6 +137,16 @@ public static class Rule
         category: Category.CodeQuality,
         isEnabled: true);
 
+    public static DiagnosticDescriptor EmbedValidResourceFiles => New(
+        id: 2000,
+        title: "Embed valid resource files",
+        message: "Invalid XML: {0}",
+        description: "A resource file should contain valid XML.",
```
LauraXiulan, line 144: Its not totally clear if it is about content or structure

rules/Proj2000.md
Corniel: I've to check explicitly which headers are required. I considered to add that (and the root node name) as a requirement, but for now this should be sufficient.

src/DotNetProjectFile.Analyzers/Rule.cs
Corniel: Because it is about both?!

LauraXiulan (approved):

PR #47, opened 2023-07-12
Use compilation bound caching

Corniel:
All rules require a context containing a representation of the project file (MS Build, or RESX) to perform their analysis. To prevent re-creating those objects over and over again for the same file, a cache has been added. This cache (with a limited capacity) uses the compilation as key, and returns the cached result or the newly created one.

Locally, for `Rules.All.Guards.DotNetProjectFile_Analyzers_project` the duration dropped from 3.1s, to 1.7s.


src/DotNetProjectFile.Analyzers/Caching/CompilationCache.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    private readonly object locker = new();
+    private readonly Compilation[] Keys;
+    private readonly T[] Values;
+    private int Next;
+
+    public CompilationCache(int capacity = 8)
+    {
+        Capacity = capacity;
+        Keys = new Compilation[8];
```
CptWesley, line 17: Shouldn't these also use the capacity?

Corniel: Euh yes...

src/DotNetProjectFile.Analyzers/MsBuild/Nodes.cs
```diff
@@ -5,28 +5,29 @@ namespace DotNetProjectFile.MsBuild;
...
     /// The parent <see cref="Node"/>.
     /// </param>
-    public Nodes(Node parent)
-    {
-        Parent = parent;
-    }
+    public Nodes(Node parent) => Items = parent.Element
+        .Elements().Select(parent.Create)
+        .Where(child => child is T)
+        .Cast<T>()
```
CptWesley, line 20: Line `19` and `20` could be replaced with `.OfType<T>()`

CptWesley (approved):

PR #48, opened 2023-07-12
New Rule Proj2003: Add invariant fallback resources

Corniel:
# Proj2003: Add invariant fallback resources
To ensure that localized values can be resolved, a localized resource file
must have a culture invariant alternative.

See: https://learn.microsoft.com/en-us/dotnet/core/extensions/localization


src/DotNetProjectFile.Analyzers/Resx/Resources.cs
```diff
@@ -14,6 +15,25 @@ internal sealed class Resources : IReadOnlyCollection<Resource>
...
+        {
+            return Array.Empty<Resource>();
+        }
+        else
+        {
+            var parents = resource.Culture.Ancestors()
+                .Select(resource.Path.Satellite)
+                .Select(file => items.TryGetValue(file, out var parent) ? parent : null)
+                .Where(r => r is { })
+                .Cast<Resource>()
```
CptWesley, line 30: Line `29` and `30` can be replaced by `.OfType<Resource>`

PR #50, opened 2023-07-14
New Rule Proj0011: Define properties once

Corniel:
# Proj0011: Define properties once
As MS Build will only select one value of a property, defining a property twice
is adding noise, if not a bug.

## Non-complaint
``` XML

  <PropertyGroup>
    <TargetFramework>net7.0</TargetFramework>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <PropertyGroup>
    <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
    <Nullable>annotations</Nullable>
  </PropertyGroup>

</Project>
```

## Compliant
``` XML

  <PropertyGroup>
    <TargetFramework>net7.0</TargetFramework>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>
```

See: #15

### DIFF
--- a/projects/PropertiesTwice/PropertiesTwice.csproj
+++ b/projects/PropertiesTwice/PropertiesTwice.csproj
@@ -6,7 +6,7 @@
     <ImplicitUsings>disable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <ProductName>Test project</ProductName>
-    <Authors>Corniel Nobel, Wesley Baartman</Authors>
+    <Authors>Corniel Nobel; Wesley Baartman</Authors>
   </PropertyGroup>
 
   <PropertyGroup>
@@ -19,7 +19,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Authors Condition="'2' == '1'">Corniel Nobel, Wesley Baartman</Authors>
+    <Authors Condition="'2' == '1'">Corniel Nobel; Wesley Baartman</Authors>
     <ProductName Condition="'1' == '1'">Test project</ProductName>
   </PropertyGroup>
 
--- a/projects/PropertiesTwice/PropertiesTwice.csproj
+++ b/projects/PropertiesTwice/PropertiesTwice.csproj
@@ -1,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <ProductName>Test project</ProductName>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFrameworks>net7.0;net6.0</TargetFrameworks>
+    <Nullable>annotations</Nullable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'1' == '1'">
+    <ProductName>Test project</ProductName>
+  </PropertyGroup>
+
+</Project>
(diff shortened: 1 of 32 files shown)


props/common.props
```diff
@@ -6,6 +6,13 @@
     <NuGetAudit>true</NuGetAudit>
   </PropertyGroup>
 
+  <PropertyGroup Label="Package">
+    <Authors>Corniel Nobel, Wesley Baartman</Authors>
```
CptWesley, line 10: `Authors` tag is delimited with semicolons (https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target).

projects/PropertiesTwice/PropertiesTwice.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <ProductName>Test project</ProductName>
+    <Authors>Corniel Nobel, Wesley Baartman</Authors>
```
CptWesley, line 9: `Authors` tag is delimited with semicolons (https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#pack-target).

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/DefinePropertiesOnce.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                context.ReportDiagnostic(Descriptor, prop, prop.LocalName);
+            }
+        }
+    }
+
+    private sealed class PropertyComparer : IEqualityComparer<Node>
+    {
+        public bool Equals(Node x, Node y)
+            => Name(x) == Name(y)
```
CptWesley, line 26: Are you sure that we there are no nodes that are expected to be defined multiple times? Or will we introduce a whitelist when we run into them in the widl?

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/DefinePropertiesOnce.cs
```diff
@@ -0,0 +1,44 @@
...
+        public int GetHashCode(Node obj)
+            => (Name(obj).GetHashCode() * 13)
+            ^ Condition(obj).GetHashCode();
+
+        private static string Name(Node node)
+            => node.LocalName == nameof(TargetFrameworks)
+            ? nameof(TargetFramework)
+            : node.LocalName;
+
+        private static string Condition(Node node)
```
CptWesley, line 38: Will this still work for nested conditions?

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/DefinePropertiesOnce.cs
Corniel: as far as I know, those do not exist. So indeed, if it turns out they exist, we can add an ingore list.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/DefinePropertiesOnce.cs
Corniel: Not yet. Given the lack of nesting, for now, I think it is sufficient.

props/common.props
```diff
@@ -6,6 +6,13 @@
     <NuGetAudit>true</NuGetAudit>
   </PropertyGroup>
 
+  <PropertyGroup Label="Package">
+    <Authors>Corniel Nobel;  Wesley Baartman</Authors>
```
CptWesley, line 10: double space

CptWesley (approved):

PR #54, opened 2023-07-15
New Rule Proj0200: Define IsPackable explicitly

CptWesley:


Corniel:
Lets make this Proj0200 so that all `<IsPacakable>` related rules can be put there.

Corniel (approved):
LGTM

PR #57, opened 2023-07-15
New Rule Proj0012: Reassign properties with different value

Corniel:
# Proj0011: Reassign properties with different value
Reassigning a value in a MS Build file only makes sense when the value differs
from the imported value.


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Reassign_properties_with_different_value.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Rules.MS_Build.Reassign_properties_with_different_value;
+
+public class Reports
+{
+    [Test]
+    public void on_reassignments_with_same_value()
+       => new ReassignPropertiesWithDifferentValue()
+       .ForProject("ReassignProperties.cs")
+       .HasIssues(
+            new Issue("Proj0012", "Property <OutputType> has previously be defined with the same value.").WithSpan(/*.*/6, 5, 6, 36),
```
CptWesley, line 10: "was previously defined" or "has previously been defined"

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ReassignPropertiesWithDifferentValue.cs
```diff
@@ -0,0 +1,39 @@
...
+                && Equals(prop.Val, previous.Val))
+            {
+                context.ReportDiagnostic(Descriptor, prop, prop.LocalName);
+            }
+        }
+    }
+
+    private static Node? EarlierAssignement(Node node, MsBuildProject project)
+    {
+        foreach (var import in project.ImportsAndSelf().Reverse().Skip(1))
```
CptWesley, line 24: This seems quite inefficient. I assume it will not be very problematic with the expected input sizes (famous last words), so if you don't think it's a pressing issue, then perhaps create an issue for this when this gets merged.

src/DotNetProjectFile.Analyzers/MsBuild/Authors.cs
```diff
@@ -1,8 +1,8 @@
 ﻿namespace DotNetProjectFile.MsBuild;
 
-public sealed class Authors : Node
+public sealed class Authors : Node<string>
 {
     public Authors(XElement element, Node parent, Project project) : base(element, parent, project) { }
 
-    public string Value => Element.Value;
+    public override string? Value => Element.Value;
```
CptWesley, line 7: Since `Element.Value` is not nullable, the `?` can be dropped. This still fits with the nullable contract of the parent. (Same applies for all other Node<string> implementations).

src/DotNetProjectFile.Analyzers/MsBuild/Node.cs
```diff
@@ -5,8 +5,18 @@
...
 namespace DotNetProjectFile.MsBuild;
 
+public abstract class Node<T> : Node
+{
+    protected Node(XElement element, Node? parent, Project? project)
+        : base(element, parent, project) { }
+
+    public sealed override object? Val => Value;
+
+    public abstract T? Value { get; }
```
CptWesley, line 15: Perhaps turn this into `public virtual T? Value => Convert<T?>(Element.Value);` To reduce the amount of copy-paste code happening in nodes that don't really have much behaviour.

Corniel: I like that suggestion.

src/DotNetProjectFile.Analyzers/MsBuild/Authors.cs
Corniel: For some reason, I was certain that it was nullable. Anyway, with the default implementation this is not longer relevant.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ReassignPropertiesWithDifferentValue.cs
Corniel: I defined a reverse too. Because I think it is better that way.

CptWesley (approved):

Issue #21, opened 2023-07-08
Support `<Choice>` nodes


Corniel:
The  `<Choice>` node should be supported.

PR #69, opened 2023-07-18
Improvement: Support choose

Corniel:
Because besides `@Condition`, also `<Choose>` containing `<When>` and `<Otherwise>` can be used to add `<PropertyGroup>`, and  `<ItemGroup>` conditionally, their children should be members of `Project.PropertyGroups` and `Propject.ItemGroups`.

See: #21

### DIFF
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -24,15 +24,16 @@
     <RepositoryUrl>https://www.github.com/Corniel/dotnet-project-file-analyzers</RepositoryUrl>
     <Description>.NET project file analyzers</Description>
     <PackageTags>Code Analysis;Project files;csproj;vbproj;resx;MS Build;resources</PackageTags>
-    <Version>1.0.7</Version>
+    <Version>1.0.8</Version>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageReleaseNotes>
-ToBeReleased
+v1.0.8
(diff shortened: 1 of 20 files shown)


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -482,7 +482,7 @@ public static class Rule
 
 #pragma warning disable S107 // Methods should not have too many parameters
     // it calls a ctor with even more arguments.
-    private static DiagnosticDescriptor New(
+    public static DiagnosticDescriptor New(
```
CptWesley, line 485: Is this intentional?

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -24,14 +24,15 @@
     <RepositoryUrl>https://www.github.com/Corniel/dotnet-project-file-analyzers</RepositoryUrl>
     <Description>.NET project file analyzers</Description>
     <PackageTags>Code Analysis;Project files;csproj;vbproj;resx;MS Build;resources</PackageTags>
-    <Version>1.0.7</Version>
+    <Version>1.0.8</Version>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageReleaseNotes>
-ToBeReleased
+v1.0.8
```
CptWesley, line 30: I think we should wait for Proj0014 (almost finished) before releasing

src/DotNetProjectFile.Analyzers/Rule.cs
Corniel: Yes. Is now used in a unit test.

PR #72, opened 2023-07-18
Improvements on handling paths

Corniel:
TODO: how to deal with variables in paths.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/IncludeProjectReferencesOnce.cs
+++ b/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/IncludeProjectReferencesOnce.cs
@@ -17,7 +17,7 @@ public sealed class IncludeProjectReferencesOnce : MsBuildProjectFileAnalyzer
             .ImportsAndSelf()
             .SelectMany(p => p.ItemGroups)
             .SelectMany(i => i.ProjectReferences)
-            .Where(r => r.Include is not null))
+            .Where(r => r.Value is not null))
         {
             var key = Reference.Create(reference);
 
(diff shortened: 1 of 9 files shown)


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/IncludeProjectReferencesOnce.cs
```diff
@@ -34,24 +35,7 @@ protected override void Register(ProjectFileAnalysisContext context)
     private record struct Reference(string Name, string Condition)
     {
         public static Reference Create(ProjectReference reference) => new(
-            Name: GetName(reference),
+            Name: FileSystem.Normalize(reference.Value!.FullName).ToLowerInvariant(),
```
LauraXiulan, line 38: Are you sure value is never null here?

src/DotNetProjectFile.Analyzers/Extensions/System.IO.FileSystemInfo.cs
```diff
@@ -11,4 +13,10 @@ public static IEnumerable<DirectoryInfo> GetAncestors(this FileInfo file)
             current = current.Parent;
         }
     }
+
+    public static FileInfo SelectFile(this DirectoryInfo directory, string path)
```
LauraXiulan, line 17: Why are SelectFile and SelectDirectory exactly the same?

Corniel: Because that is how it (should) work: you have a path, you extend that path, and it is either a file, or a directory. And that is up to developer who uses the method.

PR #73, opened 2023-07-19
Host documention at github.io

Corniel:
Use github.io to host documentation.


CptWesley:
Doesn't this imply we could get rid of the rule markdown files?

Corniel:
> Doesn't this imply we could get rid of the rule markdown files in this repository?

Yes it does.

PR #75, opened 2023-07-20
Rules should be defined a dotnet-project-file-analyzers.github.io

Corniel:
For cleaner URL's all rules defined at [dotnet-project-file-analyzers.github.io](https://dotnet-project-file-analyzers.github.io/). To prevent inconsistencies, rules should not be documented both at this GitHub repo and at dotnet-project-file-analyzers.github.io.


specs/DotNetProjectFile.Analyzers.Specs/Design_specs.cs
```diff
@@ -8,21 +9,35 @@ public class Rules
     public void Ids_are_unique()
         => Descriptors.Select(d => d.Id).Should().OnlyHaveUniqueItems();
 
-    [TestCaseSource(nameof(RuleIds))]
-    public void have_mark_down_documentation(string id)
-        => new FileInfo($"../../../../../rules/{id}.md").Exists.Should().BeTrue(because: $"{id}.md should exist.");
+    /// <remarks>
+    /// To
```
CptWesley, line 13: unfinished comment

specs/DotNetProjectFile.Analyzers.Specs/Design_specs.cs
```diff
@@ -44,18 +59,33 @@ private static IEnumerable<Type> Types
         .GetTypes()
         .Where(t => !t.IsAbstract && t.IsAssignableTo(typeof(DiagnosticAnalyzer)));
 
-    private static IEnumerable<DiagnosticDescriptor> Descriptors
+    private static IEnumerable<Rule> Descriptors
```
CptWesley, line 62: I think `Rules` might now be a more descriptive name.

specs/DotNetProjectFile.Analyzers.Specs/Design_specs.cs
Corniel: Has been dropped.

specs/DotNetProjectFile.Analyzers.Specs/Design_specs.cs
Corniel: Rules is not allowed, that's the name of the parent class, but `AllRules` is also better.

PR #88, opened 2023-08-04
New Rule Proj1200: Exclude private assets as project file dependency

Corniel:
The purpose of the rule is simple: packages that benefit from begin added as private asset should be. The best way to state this in a (positive) sentence? No clue. I'm also thinking about solutions that scale better then a static list, but do not (often) lead to false positives.


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -513,6 +513,17 @@ public static class Rule
        category: Category.Bug,
        isEnabled: true);
 
+    public static DiagnosticDescriptor SomeRule => New(
+       id: 2100,
+       title: "Add invariant fallback values",
+       message: "Mark this <PackageReference> as a private asset.",
```
CptWesley, line 519: Perhaps something like "Dependencies only used during compilation time should be marked as private assets."

CptWesley:
Isn't there some intelligent way to determine if a dependency is an analyzer? Visual studio seems capable of doing this. That way all analyzers (and by extension code generators) would automatically be identified as such.

Corniel:
> Isn't there some intelligent way to determine if a dependency is an analyzer? Visual studio seems capable of doing this. That way all analyzers (and by extension code generators) would automatically be identified as such.

I could not find how.

PR #92, opened 2023-08-15
FP/FN: Sorting of referencese should be case insensitive

Corniel:
The following should be compliant:

``` XML
 <PackageReference Include="CodeAnalysis.TestTools" Version="1.1.0" />
 <PackageReference Include="coverlet.collector" Version="*" PrivateAssets="all" />
````


CptWesley (approved):
LGTM

PR #102, opened 2024-02-20
Mark analyzer as dev dependency

### LINKED ISSUE #101: Analyzer is not marked as DevelopmentDependency
Currently the project is not marked as `DevelopmentDependency` which means that the assets will not be automatically marked as private when the dependency is installed through NuGet. Setting `DevelopmentDependency` to `true` in the `.csproj` of the analyzer should fix that.

CptWesley:
Closes #101


src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -10,6 +10,7 @@
     <RootNamespace>DotNetProjectFile</RootNamespace>
     <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <DevelopmentDependency>true</DevelopmentDependency>
```
Corniel, line 13: Side note: do we want to have a rule based on this being enabled?

Corniel (approved):
LGTM.

Issue #41, opened 2023-07-11
Investigate Directory.Build.targets


Corniel:
To include them, or not, and how to relate to it.

Corniel:
Closed with #229

PR #107, opened 2024-02-21
FP Proj0006: Directory.Build.Props is not required to be added as additional file

Corniel:
It is far from trivial to include the directory build props. Furthermore, it might be even available in all scenario's. #41


CptWesley (approved):
LGTM

PR #110, opened 2024-02-24
New rule: Build actions should have a single task

Corniel:
The nodes `<Content>`, `<Compile>`, `<None>`, `<EmbeddedResource>`, and `<AdditionalFiles>` can use `;`'s to define multiple includes/excludes/removes/updates at once. This makes things harder to read, as lines can become long.

However, should we exclude the following case? `<EmbeddedResource Include="*.txt;*.xml" />`, so if the task defines an extension?!


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Build_actions_should_have_single_task.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    [Test]
+    public void empty_nodes()
+        => new BuildActionsShouldHaveSingleTask()
+        .ForProject("MultipleBuildActionTasks.cs")
+        .HasIssues(
+            new Issue("Proj0021", "The Include 'DoesNotExist.cs' of <None> does not exist.").WithSpan(9, 5, 9, 38),
+            new Issue("Proj0021", "The Include '*.vbproj' of <Content> does not contain any files.").WithSpan(10, 5, 10, 43),
+            new Issue("Proj0021", "The Include '**/*.bin' of <EmbeddedResource> does not contain any files.").WithSpan(10, 5, 10, 43),
+            new Issue("Proj0021", "The Include '*.unknown' of <AdditionalFiles> does not contain any files.").WithSpan(12, 5, 12, 43)
```
CptWesley, line 13: These messages don't seem to match whatever is written in `Rule.cs`

CptWesley:
If possible I think allowing a task to define multiple extensions is fine. But I don't think it's very bad to enforce them to be split up.

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Build_actions_should_have_single_task.cs
Corniel: Correct. I was working on another rule first (that will be the next after this one) This was a left over, as I was not sure yet about the right way of describing this rule.

CptWesley (approved):
LGTM

PR #111, opened 2024-02-26
New Rule Proj0022: Build action includes should exist

Corniel:
The nodes `<Content>`, `<Compile>`, `<None>`, `<EmbeddedResource>`, and `<AdditionalFiles>` can have an include task. If, however, that include does not result in at least one file being added, something is wrong.


CptWesley (approved):
Nice addition!

Issue #118, opened 2024-08-12
New Rule: paths should match case sensitve


Corniel:
As most projects are compiled both on Windows (mostly locally) and UNIX (builds and/or locally), paths to the file system should match precise, so when considered case sensitive.

Corniel:
Duplicate of #71

PR #119, opened 2024-08-12
Use IOFile and IOPath to communicate with the file system

### LINKED ISSUE #117: Props not resolved on Linux with (backward) slashes
Property files (`*.props`) are not recognized/resolved by the different rules when run on Linux if the paths are Windows style with (backward) slashes.

Corniel:
There are some challenges when dealing with IO for .NET project file analyzers. First of all, Microsoft discourages (with good reason) to rely on the file system at all, assuming that all text to analyze is available to the analyzers, which is true, except for rules (like ours) that check if the state of analyzer matches the file system.

Secondly, things have to work both on UNIX and Windows systems. This makes that `\` and `/` both has to work (preferably also when mixed), and that case sensitivity should be considered a potential bug (a rule is needed here: #118). 

Finally, paths should always be fully resolved, so `/test/one/two/../../file.txt` becomes `/test/file.txt`.

As Microsoft's `System.IO.FileInfo`, `System.IO.DirectoryInfo`, and, `System.IO.Path` are powerful, but above mentioned is not all supported.

Both `IOFile` and `IOPath` are small data structures to allow access to the file system, and address those concerns. This should fix #117 .


src/DotNetProjectFile.Analyzers/Resx/Resources.cs
```diff
@@ -5,7 +5,7 @@ namespace DotNetProjectFile.Resx;
 
 public sealed class Resources : IReadOnlyCollection<Resource>
 {
-    private readonly Dictionary<FileInfo, Resource> items = new(FileSystemEqualityComparer.File);
+    private readonly Dictionary<IOFile, Resource> items = [];
```
CptWesley, line 8: maybe also use ConcurrentDictionary here?

specs/DotNetProjectFile.Analyzers.Specs/IO/IOFile_specs.cs
```diff
@@ -0,0 +1,30 @@
...
+    [TestCase("\\", @".\root\test\somefile.txt")]
+    public void Supports_formatting(string format, string formatted)
+        => IOFile.Parse("./root/test/somefile.txt").ToString(format).Should().Be(formatted);
+
+    [Test]
+    public void Is_separator_agnostic()
+        => IOFile.Parse(".\\root\\test\\somefile.txt").Should().Be(IOFile.Parse("./root/test/somefile.txt"));
+
+    [TestCase(@"c:\Program Files\Microsoft.NET")]
+    public void Supports_type_conversion(IOFile path)
```
renzo-baasdam, line 17: Is the name of this test correct? Don't see any type conversion.

renzo-baasdam (approved):

specs/DotNetProjectFile.Analyzers.Specs/IO/IOFile_specs.cs
Corniel: That the Test case works, is because of that. ;)

CptWesley (approved):
👍

Issue #91, opened 2023-08-09
Fix build


Corniel:
The build fails on running unit tests:

```
 Failed items_under_Choose [1 s]
  Error Message:
   System.InvalidOperationException : Msbuild failed when processing the file '/home/runner/work/dotnet-project-file-analyzers/dotnet-project-file-analyzers/projects/ChooseWhen/ChooseWhen.csproj' with message: /home/runner/.dotnet/sdk/7.0.400/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets: (90, 5): The "ProcessFrameworkReferences" task failed unexpectedly.
System.IO.FileLoadException: Could not load file or assembly 'NuGet.Frameworks, Version=6.7.0.111, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. Could not find or load a specific file. (0x80131621)
File name: 'NuGet.Frameworks, Version=6.7.0.111, Culture=neutral, PublicKeyToken=31bf3856ad364e35'
 ---> System.IO.FileLoadException: Could not load file or assembly 'NuGet.Frameworks, Version=6.7.0.111, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
   at System.Runtime.Loader.AssemblyLoadContext.<LoadFromPath>g____PInvoke|5_0(IntPtr ptrNativeAssemblyBinder, UInt16* ilPath, UInt16* niPath, ObjectHandleOnStack retAssembly)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
   at System.Runtime.Loader.AssemblyLoadContext.ResolveUsingLoad(AssemblyName assemblyName)
   at System.Runtime.Loader.AssemblyLoadContext.Resolve(IntPtr gchManagedAssemblyLoadContext, AssemblyName assemblyName)
   at Microsoft.NET.Build.Tasks.ProcessFrameworkReferences.<>c.<ExecuteCore>b__141_0(ITaskItem item)
   at System.Linq.Enumerable.SelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Microsoft.NET.Build.Tasks.ProcessFrameworkReferences.ExecuteCore()
[...]

PR #120, opened 2024-08-12
Drop build.sln

Corniel:
Originally, all test projects where included as projects in the main solution. As this resulted in issued at the build, a separate `build.sln` has been created. With this PR, those two solution files are merged back.

All test/reference projects are now included as items. This makes it easy to watch and/or edit them while working in the solution, but they are not build by the test projects, not by the build command.


Corniel:
Solved #91

PR #124, opened 2024-08-13
Support the suppression of rules (in XML based) files

Corniel:
I'm working on the support to suppress individual FP's of project file rules. I'm considering two options:

``` XML

  <!-- #pragma warning disable Proj0008 At parent level -->
  <ItemGroup>
    <Folder Include="First" />
  </ItemGroup>

  <ItemGroup>
    <!-- #pragma warning disable Proj0008 At own level -->
    <Folder Include="Second" />
  </ItemGroup>

  <ItemGroup>
    <Folder Include="Third" />
  </ItemGroup>

</Project>
```

Where the `#pragma warning disable` disables the next node (and its children), or more in line with C(#):

``` XML

  <!-- #pragma warning disable Proj0008 -->

  <ItemGroup>
    <Folder Include="First" />
  </ItemGroup>

  <ItemGroup>
    <Folder Include="Second" />
  </ItemGroup>

  <!-- #pragma warning restore Proj0008 -->

  <ItemGroup>
    <Folder Include="Third" />
  </ItemGroup>


  <ItemGroup>
    <AdditionalFiles Include="*.csproj" Visible="false" />
  </ItemGroup>

</Project>
```

In both cases the intended result is that `<Folder Include="Third" />` is reported on, the other two folders are ignored. Feedback is requested.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/MsBuild/PragmaWarning.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/PragmaWarning.cs
@@ -1,28 +1,46 @@
-﻿using System.Text.RegularExpressions;
+﻿using Microsoft.CodeAnalysis.Text;
+using System.Text.RegularExpressions;
+using System.Xml;
 
 namespace DotNetProjectFile.MsBuild;
 
-public readonly struct PragmaWarning(string diagnosticId, bool disable)
+/// <summary>Represents a #pragma warning (disable/restore) in a MS Build project file.</summary>
+public readonly struct PragmaWarning(string diagnosticId, bool disable, Location location)
 {
+    /// <summary>The diagnostic ID.</summary>
     public string DiagnosticId { get; } = diagnosticId;
 
-    public bool Disable { get; } = disable;
+    /// <summary>Indicates if the diagnostic is disabled (or restored).</summary>
+    public bool IsDisabled { get; } = disable;
 
+    /// <summary>The location of the #pragma warning.</summary>
+    public Location Location { get; } = location;
+
+    /// <inheritdoc />
     public override string ToString()
-        => Disable
-        ? $"#pragma warning disable {DiagnosticId}"
-        : $"#pragma resore disable {DiagnosticId}";
+        => IsDisabled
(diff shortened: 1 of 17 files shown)


LauraXiulan:
I would stick with the second one, explicitly marking the area that should be ignored. I think that is cleaner than the first one.

CptWesley:
I agree with @LauraXiulan here

Corniel:
And now I have a fully working implementation!

src/DotNetProjectFile.Analyzers/MsBuild/PragmaWarning.cs
```diff
@@ -0,0 +1,46 @@
+﻿using Microsoft.CodeAnalysis.Text;
+using System.Text.RegularExpressions;
+using System.Xml;
+
+namespace DotNetProjectFile.MsBuild;
+
+/// <summary>Represents a #pragma warning (disable/restore) in a MS Build project file.</summary>
+public readonly struct PragmaWarning(string diagnosticId, bool disable, Location location)
```
CptWesley, line 8: I'd call this a `WarningPragma`, but I understand this matches the literal syntax better.

src/DotNetProjectFile.Analyzers/MsBuild/PragmaWarning.cs
```diff
@@ -0,0 +1,46 @@
...
+        => str is { }
+        && Pattern.Match(str) is { Success: true } match
+            ? new(
+                match.Groups["DiagnosticId"].Value,
+                match.Groups["Disable"].Value == "disable",
+                location)
+            : null;
+
+    private static readonly Regex Pattern = new(
+        @"^\s*#pragma +warning +(?<Disable>disable|restore) +(?<DiagnosticId>\w+)(\s|$)",
```
CptWesley, line 43: `enable` is also a valid, but discouraged, alternative to `restore` since C#8 as a left-over of nullability analysis https://github.com/dotnet/csharplang/issues/3986

Corniel: In this context, `enable` would not work, at least not as would expect it to work: a rule that has been disabled but is enabled from a certain point and on.

PR #135, opened 2024-08-24
New Rule: XML indentation

Corniel:
XML (based filed) should be properly indented.


src/DotNetProjectFile.Analyzers/Extensions/System.Xml.Linq.XElement.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace System.Xml.Linq;
+
+internal static class XElementExtensions
+{
+    public static bool IsSelfClosing(this XElement element)
+    {
+        if (element.Value is not { Length: > 0 })
+        {
+            return element.ToString()[^2] == '/';
```
CptWesley, line 9: there can be optional spaces after the `/`, so this is not the best solution. In my PR I use a property to check for this, I don't recall by heart and doing this from phone.

Corniel: After? As far as I know, there can only be optional spacing before?!

CptWesley: Hm, it indeed appears there can not be an optional space. But using the property in place to check for this still seems like a safer bet.

CptWesley: `IsEmptyElement` on the reader of the node.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/IndentXml.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        var element = start ? node.Positions.StartElement : node.Positions.EndElement;
+
+        if (!ProperlyIndented(element))
+        {
+            var name = start ? node.LocalName : '/' + node.LocalName;
+            context.ReportDiagnostic(Descriptor, project, element, name);
+        }
+
+        bool CheckSelfClosingEnd() => !start && node.Positions.IsSelfClosing;
```
CptWesley, line 39: These 2 could just be variables, since their value will never change during the execution. I think it would make it easier to read through the method.

CptWesley (approved):
Approved with 1 comment

dotnet-project-file-analyzers.sln
```diff
@@ -52,6 +52,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "C#", "C#", "{BAD262BB-6053-
 		projects\GeneratePackageNotPackable\GeneratePackageNotPackable.csproj = projects\GeneratePackageNotPackable\GeneratePackageNotPackable.csproj
 		projects\GeneratePackageOnBuildUnconditionally\GeneratePackageOnBuildUnconditionally.csproj = projects\GeneratePackageOnBuildUnconditionally\GeneratePackageOnBuildUnconditionally.csproj
 		projects\LegacyProject\LegacyProject.csproj = projects\LegacyProject\LegacyProject.csproj
+		projects\MaliciousIndenting\MaliciousIndenting.csproj = projects\MaliciousIndenting\MaliciousIndenting.csproj
```
LauraXiulan, line 55: Malicious? Its a bit strong

Corniel: Well, do you have a better suggestion?

LauraXiulan: Faulty, improper, noncompliant or something like illegible.

LauraXiulan: Malicious feels like someone who is intent on harming with their indentation

Corniel: I went for Faulty. Thanks for the suggestion.

LauraXiulan (approved):

Issue #136, opened 2024-08-25
Location detection of XML nodes is not precise


Corniel:
(Currently) al) files validated by the .NET project file analyzers are XML. `Microsoft's System.Xml.Linq.XElement` is used as underlying data structure. To get the location information needed to report issues we depend on `XElement.Parse(string, LoadOptions.PreserveWhitespace | LoadOptions.SetLineInfo)`. This gives us, for every `XNode` parsed the start char and line.

Currenlty, to get the end position of an `XElement` (we do not report on `XNode`'s that are not `XElement`'s yet), we check if there is an next element and take its position as the end.

Let take the following project file:

``` XML
<Project Sdk="Microsoft.NET.Sdk">

  <Import Project="..\props\simple.props" />
  
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
  </PropertyGroup>

</Project>
```

In this example, if we report on the Import node, the location starts `<Import` and continues just until `<PropertyGroup`. For `<Project`, only `<` is in the selection as there is no next node, an no better effort has been made.

As Visual Studio does not show the issues with some visual indicator (Ryder and Visual Studio Code seem to do) to me personally - this had no high priority yet. Now, however, with #135 I'd like to report on the indentation it becomes relevant. Specially as we should also be able to report on the closing tag (which is not a separate node or element).

It is possible to (always) report on elements a such:

``` XML
<NodeToReportOn SomeAttribute="SomeValue">SomeText</NodeToReportOn>
 ^^^^^^^^^^^^^^
````

So, just selecting the name part of the element. This not solve the issue of detecting the closing tag however. For that, the best I think of, is using the `.ToString()` of the element, and based on that try to reverse engineer its location, combined with the `SourceText` we have. Does anyone have other suggestions?

CptWesley:
I'm not sure how this handles new lines on different operating systems and if there are other cases where they might be altering the `ToString` output.

Corniel:
@CptWesley depending on `.ToString()` would always be the last resort. Now I think of it, we might be able to use XmlReader (and not longer depend on XElement).

PR #137, opened 2024-08-26
Improvements on finding XML line position spans

Corniel:
Using the XML reader on the parsed `XElement` to determine the full span, the start element end the end element.

See: #136


Corniel:
@CptWesley I also do not think that this is the perfect solution, but it is an improvement, and it should be easier to extend on this when needed.

PR #138, opened 2024-08-27
Use AdditionalFileAnalyzerContext for AdditionalFiles

CptWesley:


specs/DotNetProjectFile.Analyzers.Specs/DotNetProjectFile.Analyzers.Specs.csproj
```diff
@@ -45,4 +45,9 @@
     <ProjectReference Include="..\..\src\DotNetProjectFile.Analyzers\DotNetProjectFile.Analyzers.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
Corniel, line 48: Can we then still test if projects are detected when they are not included as additional files? Or has this an other intended effect?

src/DotNetProjectFile.Analyzers/Extensions/Microsoft.CodeAnalysis.Diagnostics.AnalysisContext.cs
```diff
@@ -17,9 +18,22 @@ public static void RegisterProjectFileAction(this AnalysisContext context, Actio
             }
         });
```
Corniel, line 20:
```suggestion

        // Fallback for detecting files that not have been added as additional files.
```

src/DotNetProjectFile.Analyzers/MsBuild/Projects.cs
```diff
@@ -109,11 +115,20 @@ private static bool HasName(IOFile file, string? name)
     public static Projects Init(CompilationAnalysisContext context)
         => Cache.Get(context.Compilation, () => New(context));
 
+    public static Projects Init(AdditionalFileAnalysisContext context)
+        => Cache.Get(context.Compilation, () => New(context));
+
     private static Projects New(CompilationAnalysisContext context)
+        => New(context.Compilation, context.Options);
```
Corniel, line 122: This can be done simpler.

Corniel (changes requested):
I'd like the improvements. Have just some small things I'd like to see changed.

specs/DotNetProjectFile.Analyzers.Specs/DotNetProjectFile.Analyzers.Specs.csproj
CptWesley: This doesn't change this. This just adds all found csproj files to browse in the project without manually adding them one-by-one to the solution.

If we remove the CompileAction, most tests still fail (as expected).

Corniel: In that case I would go for `<None>` as we do not expect any compilation action based on in, write?

Corniel (approved):
LGTM!

PR #140, opened 2024-08-28
Proj1101: Package references should have stable versions

Corniel:
he use of nightly builds and other pre-releases should be avoided.


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Package_references_should_be_stable.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Rules.MS_Build.Package_references_should_be_stable;
+
+public class Reports
+{
+    [Test]
+    [Ignore("Buildalizer does not output has not artifacts, hence nothing is analyzed.")]
```
Sjaaky, line 6: `not output has not artifacts` ?

Sjaaky (approved):

PR #141, opened 2024-08-28
Proj2100: Indent RESX

Corniel:
Simalar to #135 , also RESX files should be properly indented.

The reason that this is a separate rule is both that it can be turned on/of independently, but also because project files and RESX files have a different analyzer context.


src/DotNetProjectFile.Analyzers/Xml/XmlAnalysisNode.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace DotNetProjectFile.Xml;
+
+public interface XmlAnalysisNode
```
CptWesley, line 3: inconsistent explicit and implicit `public` modifiers

CptWesley (approved):
1 Minor comment, otherwise LGTM

Issue #143, opened 2024-08-28
Support Central package management via Directory.Packages.props


Corniel:
As stated by Microsoft:
> To get started with central package management, you can create a `Directory.Packages.props` file at the root of your solution and set the MSBuild property ManagePackageVersionsCentrally to `true`.

So not only `Directory.Build.props`, but also `Directory.Packages.props` should be recognized as such.

More to read:
1. https://devblogs.microsoft.com/nuget/introducing-central-package-management/
2. https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management
3. https://ryanbuening.com/posts/central-package-management/
4. https://medium.com/codenx/centralize-net-project-settings-using-directory-build-props-and-directory-package-props-d6b4a471c018

PR #144, opened 2024-08-28
Introduction of ProjectFileType

Corniel:
To support scenario's like #143 (central version management) it should become possible to detect multiple types of MS-Build project files:

1. regular project files (.csproj, .vbproj)
2. regulars props files (.props)
3. Directory.Build.props
4. Directory.Version.props

This PR should facilitate this.


CptWesley (approved):
LGTM

PR #153, opened 2024-08-31
Proj0804: Only use Version when CPM is not enabled]

Corniel:
Basically the opposite of #151

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Rule.cs
+++ b/src/DotNetProjectFile.Analyzers/Rule.cs
@@ -499,7 +499,7 @@ public static class Rule
     public static DiagnosticDescriptor UseVersionOnlyWithoutCpm => New(
         id: 0804,
         title: "Use Version only with Central Package Management not enabled",
-        message: "Use VersionOverride instead of Version when CPM is enabled.",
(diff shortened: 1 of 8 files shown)


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -496,6 +496,16 @@ public static class Rule
         tags: ["Maintainability"],
         category: Category.CPM);
 
+    public static DiagnosticDescriptor UseVersionOnlyWithoutCpm => New(
+        id: 0804,
+        title: "Use Version only with Central Package Management not enabled",
+        message: "Use VersionOverride instead of Version when CPM is enabled.",
```
CptWesley, line 502: I'm not a big fan of the message here, since ideally you shouldn't need any version overrides. This makes it sound like you _need_ to use `VersionOverride` when using CPM.

CptWesley (changes requested):

src/DotNetProjectFile.Analyzers/Rule.cs
Corniel: You're right. I've changed it.

CptWesley (approved):

PR #161, opened 2024-09-02
Proj0805: could not resolve package versions in props file

### LINKED ISSUE #159: Proj0805 gives false positive with Central Package Management and Common.props file
A csproj file like:
```

  <Import Project="../../props/common.props" />
```
A common.props file like:
```
<Project>
  <ItemGroup Label="Analyzers">
    <PackageReference Include="AsyncFixer" PrivateAssets="all" />
  </ItemGroup>
</Project>

```
And a Directory.Packages.Props file like:
```
<Project>
  <PropertyGroup>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
    <CentralPackageFloatingVersionsEnabled>true</CentralPackageFloatingVersionsEnabled>
  </PropertyGroup>
  <ItemGroup>
    <PackageVersion Include="AsyncFixer" Version="*-*" />
  </ItemGroup>
</Project>

```

Returns the warning:
common.props: warning Proj0805: Define version for 'AsyncFixer' PackageReference. (https://dotnet-project-file-analyzers.github.io/rules/Proj0805.html)

Corniel:
It turns out that the `Directory.Build.props` and `Directory.Packages.props` where not linked to `*.props` when they should. Another issue with this rule is due to the fact that MSB4018 fails with a crash. It seems that `Proj0805` therefor is not hit, and does not give the info we need. :(

Closes #159

### DIFF
--- a/projects/PackageReferenceWithoutVersion/Directory.Packages.props
+++ b/projects/PackageReferenceWithoutVersion/Directory.Packages.props
@@ -1,0 +1,17 @@
+<Project>
+
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="Qowaiv" Version="7.0.0" />
+    <PackageVersion Include="Qowaiv.Analyzers.CSharp" Version="2.0.2" />
+  </ItemGroup>
+  
+  <ItemGroup>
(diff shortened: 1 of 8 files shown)


CptWesley:
Why change the entire test setup rather than adding a new test case? `PackageVersion` can be used without central package management and this test was also testing that use case (which seemed to work)

Corniel:
Can `PackageVersion` be used without CPM? If I opened the test project, I saw that it resolved Qowaiv with version `1.0.139`, the first every released, and not with the version defined in the `<PackageVersion>`.

PR #162, opened 2024-09-03
Proj1702: Omit XML declarations

Corniel:
The XML declaration is redundant for MS Build project files.


LauraXiulan (approved):

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Ommit_XML_declaration.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Rules.MS_Build.Ommit_XML_declaration;
```
LauraXiulan, line 1:
```suggestion
﻿namespace Rules.MS_Build.Omit_XML_declaration;
```
Also in the filename

PR #166, opened 2024-09-03
Fix old urls in package info

CptWesley:


Corniel (approved):
Oops. Thanks for fixing it!

PR #172, opened 2024-09-03
<IsTestProject> is true ensure Proj0200 and Proj0400 do not report

### LINKED ISSUE #168: FP Proj0200 & Proj0400: When <IsTestProject> is true is should be obvious
If `<IsTestProject>true</IsTestProject>` it should be obvious that the project is packable, nor publishable.

Corniel:
Closes #168


CptWesley:
I wouldn't close the attached issue unless a new issue is opened to deal with the 2 issues mentioned in my last comment in the issue

CptWesley (approved):

PR #174, opened 2024-09-04
Proj0014: Ignore <ProjectReference ReferenceOutputAssembly="false">

### LINKED ISSUE #169: FP Proj0014: a double reference should not count for ReferenceOutputAssembly="false"
In our own test project - if we add the main project also as analyzer - that should not count as a double project reference.

Corniel:
Closes #169


CptWesley (changes requested):
-

Corniel:
I guess you're right. I'll close this issue.

PR #177, opened 2024-09-04
Proj0450 & Proj0451: Test project should publish nor package

Corniel:
Packabillity and Publishabillity should be handled differently for test projects in that sense that they should always be set to false.


projects/PackableTestProject/PackableTestProject.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
LauraXiulan, line 1: This and the implicit project are both for Packable and Publishable. Can you either separate the two or make it clear in the project name.

specs/DotNetProjectFile.Analyzers.Specs/DotNetProjectFile.Analyzers.Specs.csproj
```diff
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
     <IsTestProject>true</IsTestProject>
+    <IsPackable>true</IsPackable>
```
LauraXiulan, line 8: Why is this added here?

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -447,6 +447,26 @@ public static class Rule
         tags: ["Configuration", "confusion"],
         category: Category.Clarity);
 
+    public static DiagnosticDescriptor TestProjectShouldNotBePackable => New(
+        id: 0450,
+        title: "Test projects should not be packable",
+        message: "Set <IsPackable> to false.",
+        description:
+            "Test projects should only be responsible for running tests. Hence " +
+            "the should not be packable.",
```
LauraXiulan, line 456:
```suggestion
            "they should not be packable.",
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -447,6 +447,26 @@ public static class Rule
...
+        tags: ["Configuration"],
+        category: Category.Bug);
+
+    public static DiagnosticDescriptor TestProjectShouldNotBePublishable => New(
+        id: 0451,
+        title: "Test projects should not be publishable",
+        message: "Set <IsPublishable> to false.",
+        description:
+            "Test projects should only be responsible for running tests. Hence " +
+            "the should not be publishable.",
```
LauraXiulan, line 466:
```suggestion
            "they should not be publishable.",
```

LauraXiulan (changes requested):

projects/PackableTestProject/PackableTestProject.csproj
Corniel: Suggestion for the name?

specs/DotNetProjectFile.Analyzers.Specs/DotNetProjectFile.Analyzers.Specs.csproj
Corniel: Oops.

projects/PackableTestProject/PackableTestProject.csproj
LauraXiulan: PackableAndPublishableTestProject

LauraXiulan: Or without the 'And'

CptWesley (approved):

PR #196, opened 2024-09-12
Use walker to navigate to MS Build file tree

### LINKED ISSUE #194: Issues reported on wrong file can lead to crashing analyzers (AD0001)
A colleague of mine reported the following error:

> CSC : warning AD0001: Analyzer 'DotNetProjectFile.Analyzers.MsBuild.OrderUsingDirectivesAlphabetically' threw an exception of type 'System.ArgumentOutOfRangeException' with message 'Specified argument was out of the range of valid values.

Some local debugging showed what was going wrong: Some rules 

``` C#
protected override void Register(ProjectFileAnalysisContext context)
{
    foreach (var references in context.Project
        .ImportsAndSelf()
//       ^^^^^^^^^^^^^^^^
        .SelectMany(p => p.ItemGroups)
        .Select(g => g.PackageReferences))
    {
        AnalyzeGroup(context, references);
    }
}
```

The problem with this construct is that when reported on, the context is original context is used. That means if an issue is found on any of the imports, its reported on the ancestor instead, which is wrong. However, if the span of the reported node is outside the `TextSource` of the ancestor context it just crashes with the mentioned `OutOfRangeException`.

For some of the analyzers, just only analyzing the own nodes will work like a charm, the imports are analyzed as well anyway. However, for some rules, the issues found in the imports are found only because of the analyzed ancestor, in those cases we have to reconsider the mechanism of reporting.

Corniel:
A walker better reflects how values are determined by MS Build.

Closes #194


src/DotNetProjectFile.Analyzers/MsBuild/Project.cs
```diff
@@ -97,30 +81,21 @@ public IEnumerable<Project> ImportsAndSelf()
         yield return this;
     }
 
-    /// <summary>Loops through all imports and self reversed.</summary>
-    public IEnumerable<Project> SelfAndImports()
+    /// <summary>Gets self, Directory.Packages.props, and Directory.Build.props).</summary>
+    /// <remarks>
+    /// If the *.props are null, or higher in the type hierarchy they are skipped. 
+    /// </remarks>
+    private IEnumerable<Project> SelftAndDirectoryProps()
```
CptWesley, line 88: `SelfAndDirectoryProps`

CptWesley (approved):

src/DotNetProjectFile.Analyzers/MsBuild/Project.Property.cs
```diff
@@ -0,0 +1,76 @@
...
+    private bool? managePackageVersionsCentrally;
+
+    public TNode? Property<TNode>() where TNode : Node => this
+        .SelfAndDirectoryProps()
+        .Select(p => Read<TNode>(p, new(p.Path)))
+        .OfType<TNode>()
+        .FirstOrDefault();
+
+    /// <remarks>
+    /// Walk in reversed order true the nodes.
```
Sjaaky, line 48: through

Issue #201, opened 2024-09-20
Support .editorconfig Glob Expressions


Corniel:
To fully support `.editorconfig` files, not ony the parsing of an INI file (#199 ) is required, but also the support of [Glob Expressions](https://spec.editorconfig.org/index.html#glob-expressions).

| Special Characters | Matching|
|--------------------|----------------------
| `*`                | any string of characters, except path separators (/)
| `**`               | any string of characters
| `?`                | any single character, except path separators (/)
| `[seq]`            | any single character in seq
| `[!seq]`           | any single character not in seq
| `{s1,s2,s3}`       | any of the strings given (separated by commas, can be nested) (But {s1} only matches {s1} literally.)
| `{num1..num2}`     | any integer numbers between num1 and num2, where num1 and num2 can be either positive or negative

Without that, we can not determine which sections are relevant for specific files.

sungam3r:
@Corniel Am I understand correctly that now I can’t configure the suppression of diagnostics through the editorconfig file? Like

```
[*.{props,csproj}]
dotnet_diagnostic.Proj0003.severity = suggestion
```

Corniel:
> @Corniel Am I understand correctly that now I can’t configure the suppression of diagnostics through the editorconfig file? Like
> 
> ```
> [*.{props,csproj}]
> dotnet_diagnostic.Proj0003.severity = suggestion
> ```

Nope, you can't. That's because how MS Build works.  See [configuration](https://dotnet-project-file-analyzers.github.io/configuration.html) documentation.

Frulfump:
> > [@Corniel](https://github.com/Corniel) Am I understand correctly that now I can’t configure the suppression of diagnostics through the editorconfig file? Like
> > ```
> > [*.{props,csproj}]
> > dotnet_diagnostic.Proj0003.severity = suggestion
> > ```
> 
> Nope, you can't. That's because how MS Build works. See [configuration](https://dotnet-project-file-analyzers.github.io/general/configuration.html) documentation.

The link gives me a 404

Corniel:
I've updated the URL: https://dotnet-project-file-analyzers.github.io/configuration.html

PR #202, opened 2024-09-21
Implement glob expressions

Corniel:
To support `.editorconfig` files we need to support glob expressions (see #201). This implementation converts to `Regex`'s (for simplicity). There is at least one dedicated [.NET implementation](https://github.com/kthompson/glob) but the downside would be to have an external dependency on that implementation, or to have to include/duplicate a lot of code, compared to what was needed to have a working implementation.


Corniel:
Done differently with #207

PR #200, opened 2024-09-20
Define file cache that is shared amongst all analyzers

Corniel:
To improve in speed, and to have an easier to follow path for retrieving (parsed) files.

| Project      |    NEW     |    OLD     |
|--------------|-----------:|-----------:|
| Compliant CS |   7.807 ms |  28.173 ms |
| Compliant VB |   7.737 ms |  27.440 ms |
| Own project  |  69.195 ms | 313.239 ms |

PR #204, opened 2024-09-23
Support rules on .editorconfig files

Corniel:
After some refactorings (#200, #203) is is now possible to implement rules on INI-files and/or `.editorconfig` files. As it turned out that Microsoft also has a [glob implementation](https://learn.microsoft.com/dotnet/core/extensions/file-globbing] I thought, it might be possible to use that dependency (less likely to result in version conflicts for users of our analyzers).

While implementing this rule us MS's implementation, it seems that they accept just everything. :(


CptWesley (approved):

Corniel:
@CptWesley With this code we can not invalidate globs. I think that is problematic. What do you think?

CptWesley:
> @CptWesley With this code we can not invalidate globs. I think that is problematic. What do you think?

What do you mean?

Corniel:
Well, by using Microsoft's implementation `*.{exe,bin}}` will not be invalidated, although that is not a valid glob. Other external globs might do that, but I do not want to depend on them. I've an own implementation (#202 ) But is rather slow.

CptWesley:
Shouldn't we validate the same way that microsoft uses them?

Corniel:
> Shouldn't we validate the same way that microsoft uses them?

Well, when deciding how to deal with the values: yes. But when we want to report on issues, we might want to be stricter. But for now, I just removed the rule itself, and *only* provided the framework to define rules, and to be able to read `.editorconfig` files in project files.

CptWesley:
The general editorconfig specs (https://spec.editorconfig.org/) do allow those kind of patterns. I would say that it makes sense to follow the official specification of whatever file format we are dealing with, rather than make it more strict because other tools can't handle it.

Corniel:
I think is not about being able to deal with it, but if someone types: `*.{csproj,vbproj}}` it is most likely, that  `*.{csproj,vbproj}` was intended, instead of matching on either `*.csproj` or `*.vbproj}`. Anyhow, food for thought for PR's to follow.

CptWesley:
Ah I didn't see the double `}`, I thought you wanted to ban `{}`-like expressions because they are not supported by unix shell glob searches

PR #210, opened 2024-10-01
Proj3000: Only use UTF-8 without BOM

Corniel:
There are at least three problems with putting a BOM in UTF-8 encoded files.

1. Files that hold no text are no longer empty because they always contain the BOM.
2. Files that hold text within the ASCII subset of UTF-8 are no longer themselves ASCII because the BOM is not ASCII, which makes some existing tools break down, and it can be impossible for users to replace such legacy tools.
3. It is not possible to concatenate several files together because each file now has a BOM at the beginning.

Source: https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-with-bom

There are multiple other sources stating basically the same.

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/Generic/Only_use_UTF-8_without_BOM.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/Generic/Only_use_UTF-8_without_BOM.cs
@@ -1,4 +1,4 @@
-using DotNetProjectFile.Analyzers.Files;
+using DotNetProjectFile.Analyzers.Generic;
 
 namespace Rules.Generic.Only_use_UTF8_without_BOM;
 
--- a/src/DotNetProjectFile.Analyzers/Rule.cs
+++ b/src/DotNetProjectFile.Analyzers/Rule.cs
@@ -768,13 +768,12 @@ public static class Rule
     public static DiagnosticDescriptor OnlyUseUTF8WithoutBom => New(
         id: 3000,
         title: "Ony use UTF-8 encoding without BOM",
-        message: "This file's encoding is using UTF-8 with BOM.",
+        message: "This file is using UTF-8 encoding with BOM.",
         description:
             "The use of BOM for UTF-8 can cause systems to malfunction. This " +
             "includes multiple web based files such as Javascript and CSS.",
         tags: ["encoding", "UTF-8"],
         category: Category.Reliability);
-
 
 #pragma warning disable S107 // Methods should not have too many parameters
     // it calls a ctor with even more arguments.
--- a/specs/Benchmarks/Benchmarks.csproj
+++ b/specs/Benchmarks/Benchmarks.csproj
@@ -6,6 +6,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <IsPackable>false</IsPackable>
     <OutputType>exe</OutputType>
+    <EnforceExtendedAnalyzerRules>true</EnforceExtendedAnalyzerRules>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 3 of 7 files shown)


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -765,6 +765,17 @@ public static class Rule
       tags: ["XML", "indentation"],
       category: Category.Formatting);
 
+    public static DiagnosticDescriptor OnlyUseUTF8WithoutBom => New(
+        id: 3000,
+        title: "Ony use UTF-8 encoding without BOM",
+        message: "This file's encoding using UTF-8 with BOM.",
```
CptWesley, line 771: this message does not contain a verb

src/DotNetProjectFile.Analyzers/Analyzers/Files/OnlyUseUTF8WithoutBom.cs
```diff
@@ -0,0 +1,46 @@
...
+public sealed class OnlyUseUTF8WithoutBom() : ProjectFileAnalyzer<ProjectTextFile>(Rule.OnlyUseUTF8WithoutBom)
+{
+    protected override void Register(AnalysisContext context)
+    {
+        context.RegisterCompilationAction(c =>
+        {
+            if (ProjectFiles.Global.UpdateMsBuildProject(c) is { } msbuild)
+            {
+                foreach (var file in msbuild.Path.Directory.Files("**/*")?
+                    .Where(f => !f.ToString("/").Contains("/bin/"))
```
CptWesley, line 15: Just a sanity check, but is is this always `/bin/` on all operating systems?

CptWesley:
Looks good besides my comments

src/DotNetProjectFile.Analyzers/Rule.cs
LauraXiulan, line 771:
```suggestion
        message: "This file's encoding is using UTF-8 with BOM.",
```

src/DotNetProjectFile.Analyzers/Analyzers/Files/OnlyUseUTF8WithoutBom.cs
Corniel: See the `.ToString("/")`. This is 'simple' implementation. I'm working on a more robust one.

specs/DotNetProjectFile.Analyzers.Specs/Rules/Generic/Only_use_UTF-8_without_BOM.cs
```diff
@@ -0,0 +1,24 @@
...
+
+public class Reports
+{
+    [Test]
+    public void files_with_UTF8_BOM()
+       => new OnlyUseUTF8WithoutBom()
+       .ForProject("MultipleEncodings.cs")
+       .HasIssues(
+           new Issue("Proj3000", "This file's encoding using UTF-8 with BOM.").WithPath("utf8-bom.css"),
+           new Issue("Proj3000", "This file's encoding using UTF-8 with BOM.").WithPath("file.utf8-bom"));
```
LauraXiulan, line 13:
```suggestion
           new Issue("Proj3000", "This file's encoding is using UTF-8 with BOM.").WithPath("utf8-bom.css"),
           new Issue("Proj3000", "This file's encoding is using UTF-8 with BOM.").WithPath("file.utf8-bom"));
```

specs/DotNetProjectFile.Analyzers.Specs/Rules/Generic/Only_use_UTF-8_without_BOM.cs
```diff
@@ -0,0 +1,24 @@
...
+}
+
+public class Guards
+{
+    [TestCase("CompliantCSharp.cs")]
```
LauraXiulan, line 18: Can you add an explicit utf8 without bom encoded file to this project

Corniel: The project files themselves are using UTF-8 without BOM.

PR #217, opened 2024-11-01
XML nodes within resource values should be escaped

### LINKED ISSUE #215: The `<data>` node in a resource file should only contain text
As triggered by #214 , the  `<data>` element in a resource file, should only contain text. So (HTML) nodes should be escaped or captured in `<[CDATA[ ]]>`.

Corniel:
To ensure correct handling, XML nodes within resource values should be escaped.

Closes #215 

Documentation PR: https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers.github.io/pull/42

### DIFF
--- a/projects/ResxWithHtmlContent/Messages.resx
+++ b/projects/ResxWithHtmlContent/Messages.resx
@@ -1,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <resheader name="resmimetype">
+    <value>text/microsoft-resx</value>
+  </resheader>
+   <resheader name="reader">
+    <value>System.Resources.ResXResourceReader</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter</value>
+  </resheader>
+  <data name="Html" xml:space="preserve">
+    <value><b>King</b></value>
+  </data>
+  <data name="Escaped" xml:space="preserve">
+    <value>&lt;b&gt;King&lt;/b&gt;</value>
+  </data>
+  <data name="Cdata" xml:space="preserve">
+     <value><![CDATA[<b>King</b>]]></value>
+  </data>
+</root>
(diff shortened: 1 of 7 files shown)


projects/ResxWithHtmlContent/Messages.resx
```diff
@@ -0,0 +1,21 @@
...
+  <resheader name="resmimetype">
+    <value>text/microsoft-resx</value>
+  </resheader>
+   <resheader name="reader">
+    <value>System.Resources.ResXResourceReader</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter</value>
+  </resheader>
+  <data name="Html" xml:space="preserve">
```
LauraXiulan, line 12: Shouldn't this node and the two nodes below in separate projects so you can show one 'fails' and the other 'succeeds'. The CompliantCSharp package doesn't have html content.

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -757,6 +757,14 @@ public static class Rule
        tags: ["resx", "resources", "invariant", "localization"],
        category: Category.Bug);
 
+    public static DiagnosticDescriptor EscapeXmlNodesResourceValues => New(
+      id: 2005,
+      title: "Escape XML nodes of resource values",
+      message: "Escape the XML node in '{0}'.",
+      description: "To ensure correct handling, XML nodes within resource values should be escaped.",
```
LauraXiulan, line 764: Just for clarification, is it XML nodes or HTML nodes that need to be escaped?

Corniel: Most likely the XML in these data values are HTML, but technically, all XML inside should be escaped. Hence the choice of words.

projects/ResxWithHtmlContent/Messages.resx
Corniel: I would say not. The fact that is reports on the non-compliant, and not an the escaped compliant nodes seems fine to me.

LauraXiulan (approved):

PR #221, opened 2024-11-10
New Rule: Node might be misspelled

### LINKED ISSUE #220: New Rule: Report on unkown nodes
Misspelled XML nodes/elements can lead to hard to track behavior. Tracking those unknown nodes is easy:

``` C#
protected override void Register(ProjectFileAnalysisContext<MsBuildProject> context)
{
    Walk(context.File, context);
}

private void Walk(Node node, ProjectFileAnalysisContext<MsBuildProject> context)
{
    if (node is Unknown)
    {
        context.ReportDiagnostic(Descriptor, node, node.LocalName);
    }

    foreach (var child in node.Children)
    {
        Walk(child, context);
    }
}
```

reporting on a node that is similar to a known node (hence a spelling/casing error) is also doable, and would lead to a rule with - at least I think - a lot of added value. 

Unfortunately, it is less straightforward to get a collection of supported nodes. A rule like this should allow to specify known elements, but still, without adding a lot of known nodes first, creating a rule like this would lead to a lot FP's. Any thoughts?

Corniel:
Naming of the rule and the related documentation is not finalized. It also could be that this should be two rules: one that reports on unknown nodes that have a small Levenshtein distance to a known one, and one for those who don't, as they are more likely False Positives.

This would allow users to disable that second rule (hopefully after reporting those issues), and keep the rule that reports potential miss spelled nodes enabled, as it less likely to report on FP's.

Closes #220


specs/DotNetProjectFile.Analyzers.Specs/DotNetProjectFile.Analyzers.Specs.csproj
```diff
@@ -18,6 +18,8 @@
 
   <ItemGroup Label="Project under test">
     <ProjectReference Include="../../src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj" />
+    <ProjectReference Include="../DotNetProjectFile.Analyzers.TestTools\DotNetProjectFile.Analyzers.TestTools.csproj" />
```
CptWesley, line 21: contains backwards slash

src/DotNetProjectFile.Analyzers/Text/Levenshtein.cs
```diff
@@ -0,0 +1,82 @@
+namespace DotNetProjectFile.Text;
+
+/// <summary>
+/// Measures the difference between two strings.
+/// Uses the Levenshtein string difference algorithm.
+///
+/// See: https://en.wikipedia.org/wiki/Levenshtein_distance.
+/// </summary>
+/// <remarks>
+/// Copyright: Dan Hartley, https://github.com/DanHarltey/Fastenshtein.
```
CptWesley, line 10: MIT license requires a reproduction of the license text with any reproduction of the code, so I would also include that here in the remarks.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/NewRule.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        var name = node.LocalName.ToUpperInvariant();
+        string? suggestion = null;
+        var best = name.Length / 2;
+
+        var word = new Levenshtein(name);
+
+        foreach (var known in Knowns)
+        {
+            var test = word.DistanceFrom(known);
```
CptWesley, line 38: Just as a general sanity check: are csproj node types case sensitive? Is this levenshtein implementation case sensitive?

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/NewRule.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            Walk(child, context);
+        }
+    }
+
+    private string? Suggestion(Node node)
+    {
+        var name = node.LocalName.ToUpperInvariant();
+        string? suggestion = null;
+        var best = name.Length / 2;
```
CptWesley, line 32: Maybe also good to document this minimum difference before being reported in the documentation somewhere.

src/DotNetProjectFile.Analyzers/MsBuild/OutputType.cs
```diff
@@ -1,5 +1,17 @@
 namespace DotNetProjectFile.MsBuild;
 
+/// <summary>
+/// Specifies the file format of the output file. This parameter can have one
+/// of the following values:
+/// - Library: Creates a code library. (Default value.)
+/// - Exe:    Creates a console application.
+/// - Module: Creates a module.
+/// - Winexe: Creates a Windows-based program.
```
CptWesley, line 9: Isn't this supposed to be `WinExe` rather than `Winexe`?

src/DotNetProjectFile.Analyzers/Text/Levenshtein.cs
```diff
@@ -0,0 +1,82 @@
...
+                costs[j] = curr;
+                prev = curr;
+            }
+        }
+        return costs[^1];
+    }
+
+    /// <inheritdoc />
+    [Pure]
+    public override string ToString() => Value ?? string.Empty;
```
CptWesley, line 81: I find this quite strange `ToString`, but it's not very important I think.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/NewRule.cs
```diff
@@ -0,0 +1,47 @@
+using DotNetProjectFile.Text;
+
+namespace DotNetProjectFile.Analyzers.MsBuild;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+public sealed class NewRule() : MsBuildProjectFileAnalyzer(Rule.NewRule)
```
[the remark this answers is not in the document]
Corniel: I agree: have no clue yet which one. Do you have a suggestion?

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/NewRule.cs
Corniel: Nodes are case-sensitive. I test case-insensitive. This ensures that a node with only different casing is always the best suggestion.

src/DotNetProjectFile.Analyzers/MsBuild/OutputType.cs
Corniel: Those values are not case-sensitive, and just copied from MS. I'll check the casing though.

src/DotNetProjectFile.Analyzers/Text/Levenshtein.cs
Corniel: I Can also put this in a debugger display, that might make more sense.

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Correct_spelling_of_nodes.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Rules.MS_Build.Correct_spelling_of_nodes;
+
+public class Reports
+{
+    [Test]
+    public void on_spelling_errors() => new CorrectSpellingOfNodes()
+        .ForProject("Misspelled.cs")
+        .HasIssues(
+            new Issue("Proj0031", "The node name EnableNetAnalyzers is misspelled, EnableNETAnalyzers might be intended."),
```
CptWesley, line 12: I feel like this message might be a bit too aggressive, I would maybe go for something along the lines of how Microsoft Office Word phrases it

PR #223, opened 2024-11-11
New Rule: Use VB.NET specific properties only in when applicable

Corniel:
When working on #221 I was reminded that there are properties that are only applicable in certain contexts. Hence this rule.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Rule.cs
+++ b/src/DotNetProjectFile.Analyzers/Rule.cs
@@ -257,7 +257,7 @@ public static class Rule
     public static DiagnosticDescriptor UseInVBContextOnly => New(
         id: 0030,
         title: "Use VB.NET specific properties only in when applicable",
-        message: "The property {0} is only applicable when using VB.NET and can therefor be removed.",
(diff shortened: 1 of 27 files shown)


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -254,6 +254,17 @@ public static class Rule
         tags: ["conditions"],
         category: Category.Reliability);
 
+    public static DiagnosticDescriptor UseInVBContextOnly => New(
+        id: 0030,
+        title: "Use VB.NET specific properties only when applicable",
+        message: "The property <{0}> is only applicable when using VB.NET and can therefor be removed.",
+        description:
+            "Properties only applicable to VB.NET are noise when none of " +
+            "involved targets is a VB.NET target.",
```
LauraXiulan, line 263:
```suggestion
            "Properties only applicable to VB.NET are noise when none of " +
            "the involved targets is a VB.NET target.",
```

LauraXiulan (approved):

PR #224, opened 2024-11-11
Proj0029: Use C# specific properties only when applicable

Corniel:
Equivalent to #223, but for C#. Currently, only 3 properties are involved.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseInCSharpContextOnly.cs
+++ b/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseInCSharpContextOnly.cs
@@ -1,0 +1,42 @@
+namespace DotNetProjectFile.Analyzers.MsBuild;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
+public sealed class UseInCSharpContextOnly() : MsBuildProjectFileAnalyzer(Rule.UseInCSharpContextOnly)
+{
+    private static readonly Type[] CSharpOnly =
+    [
+        typeof(DotNetProjectFile.MsBuild.CSharp.AllowUnsafeBlocks),
+        typeof(DotNetProjectFile.MsBuild.CSharp.CheckForOverflowUnderflow),
+        typeof(DotNetProjectFile.MsBuild.CSharp.Nullable),
+    ];
+
+    protected override IReadOnlyCollection<ProjectFileType> ApplicableTo => ProjectFileTypes.ProjectFile_DirectoryBuild;
+
+    protected override void Register(ProjectFileAnalysisContext<MsBuildProject> context)
+    {
+        if (!InCSharpContext(context.File))
+        {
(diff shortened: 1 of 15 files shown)


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseInCSharpContextOnly.cs
```diff
@@ -0,0 +1,42 @@
...
+        if (!InCSharpContext(context.File))
+        {
+            Walk(context.File, context);
+        }
+    }
+
+    private static bool InCSharpContext(MsBuildProject project) => project.FileType switch
+    {
+        ProjectFileType.ProjectFile => project.Path.Extension.IsMatch(".csproj"),
+        ProjectFileType.DirectoryBuild => !project.Path.Directory.Files("*.vbproj").Any(),
```
LauraXiulan, line 26: In the vbonly rule you check if there are any vbproj files, shouldn't you check here if there are any csproj files

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -254,6 +254,16 @@ public static class Rule
         tags: ["conditions"],
         category: Category.Reliability);
 
+    public static DiagnosticDescriptor UseInCSharpContextOnly => New(
+        id: 0029,
+        title: "Use C# specific properties only when applicable",
+        message: "The property <{0}> is only applicable when using C# and can therefor be removed.",
+        description:
+            "Properties only applicable to VB.NET are noise when none of " +
+            "the involved targets is a VB.NET target.",
```
LauraXiulan, line 263:
```suggestion
            "Properties only applicable to C# are noise when none of " +
            "the involved targets is a C# target.",
```

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseInCSharpContextOnly.cs
Corniel: No. There are more extension types which can be other languages, but also other C# based types, where there is (as far as I know) only one type of project file extension linked to VB.NET. So hence the asymmetrical implementation.

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -28,6 +28,7 @@
     <PackageReleaseNotes>
       <![CDATA[
 ToBeReleased
+- Proj0029: Use C# specific properties only when applicable. (NEW RULE)
```
Gemberkoekje, line 31: Should the version be updated?

Gemberkoekje (approved):

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
Corniel: Not yet.

PR #228, opened 2024-11-12
Proj0031: Adopt preferred casing of nodes

Corniel:
After some consideration, I think it is better to first release a rule just checking for consistent casing, instead of also checking for potential spelling issues (#221). This should not (hardly) lead to any FP's, where the other might bring more harm then good (hopefully not).

### DIFF
--- a/projects/CaseInsensitive/CaseInsensitive.csproj
+++ b/projects/CaseInsensitive/CaseInsensitive.csproj
@@ -1,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\props\simple.props" />
+  
+  <PropertyGroup>
+    <TARGETFRAMEWORK>net8.0</TARGETFRAMEWORK>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ComPile Include="../common/Code.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <packagereference Include="Qowaiv" Version="7.1.3" />
+  </ItemGroup>
+  
+</Project>
--- a/projects/CaseInsensitive/CaseInsensitive.csproj
+++ b/projects/CaseInsensitive/CaseInsensitive.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="..\props\simple.props" />
+  <Import Project="../props/simple.props" />
   
   <PropertyGroup>
     <TARGETFRAMEWORK>net8.0</TARGETFRAMEWORK>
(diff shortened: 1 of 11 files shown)


src/DotNetProjectFile.Analyzers/MsBuild/NodeFactory.cs
```diff
@@ -8,43 +8,58 @@ namespace DotNetProjectFile.MsBuild;
...
+    public Node Create(XElement element, Node parent, MsBuildProject project)
+        => Lookup(element).TryGetValue(element.Name.LocalName, out var con)
+        ? con(element, parent, project)
+        : new Unknown(element, parent, project);
+
+    [Pure]
+    private Dictionary<string, CtorFunc> Lookup(XElement element)
+        => element.Depth() >= 2
+        ? CaseInsensitve
+        : CaseSensitve;
```
LauraXiulan, line 36:
```suggestion
    public NodeFactory()
    {
        CtorArgumentTypes = GetCtorParameterTypes();
        CtorArgumentExpressions = CtorArgumentTypes.Select(x => Expression.Parameter(x)).ToArray();
        CaseSensitive = BuildCtorMap();
        CaseInsensitive = new Dictionary<string, CtorFunc>(CaseSensitive, StringComparer.OrdinalIgnoreCase);
    }

    private readonly Type[] CtorArgumentTypes;
    private readonly ParameterExpression[] CtorArgumentExpressions;
    private readonly Dictionary<string, CtorFunc> CaseSensitive;
    private readonly Dictionary<string, CtorFunc> CaseInsensitive;

    [Pure]
    public Node Create(XElement element, Node parent, MsBuildProject project)
        => Lookup(element).TryGetValue(element.Name.LocalName, out var con)
        ? con(element, parent, project)
        : new Unknown(element, parent, project);

    [Pure]
    private Dictionary<string, CtorFunc> Lookup(XElement element)
        => element.Depth() >= 2
        ? CaseInsensitive
        : CaseSensitive;
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -274,6 +274,16 @@ public static class Rule
         tags: ["noise"],
         category: Category.Noise);
 
+    public static DiagnosticDescriptor AdoptPreferredCasing => New(
+        id: 0031,
+        title: "Adopt preferred casing of nodes",
+        message: "The node <{0}> is has a different casing than the preferred one <{1}>.",
```
LauraXiulan, line 280:
```suggestion
        message: "The node <{0}> has a different casing than the preferred one <{1}>.",
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -274,6 +274,16 @@ public static class Rule
...
+        description:
+            "MS Build is (mostly) case insensitive. To prevent issues, however " +
+            "it is  preferred to use the same casing consistently.",
```
LauraXiulan, line 283:
```suggestion
            "MS Build is (mostly) case insensitive. To prevent issues, however, " +
            "it is preferred to use the same casing consistently.",
```

projects/CaseInsensitive/CaseInsensitive.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\props\simple.props" />
```
LauraXiulan, line 3:
```suggestion
  <Import Project="../props/simple.props" />
```

LauraXiulan (approved):

PR #235, opened 2024-11-15
 Proj3000: Inaccessible files should not lead to a crash

Corniel:
Just be defensive when accessing a file.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Analyzers/Generic/OnlyUseUTF8WithoutBom.cs
+++ b/src/DotNetProjectFile.Analyzers/Analyzers/Generic/OnlyUseUTF8WithoutBom.cs
@@ -32,7 +32,7 @@ public sealed class OnlyUseUTF8WithoutBom() : ProjectFileAnalyzer<ProjectTextFil
     {
         if (!context.File.Path.Exists) { return; }
 
-        using var reader = context.File.Path.OpenRead();
+        using var reader = context.File.Path.TryOpenRead();
 
         var bom = new byte[3];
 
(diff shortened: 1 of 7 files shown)


src/DotNetProjectFile.Analyzers/Analyzers/Generic/OnlyUseUTF8WithoutBom.cs
```diff
@@ -32,7 +32,7 @@ protected override void Register(ProjectFileAnalysisContext<ProjectTextFile> con
     {
         if (!context.File.Path.Exists) { return; }
 
-        using var reader = context.File.Path.OpenRead();
+        using var reader = context.File.Path.TryOpenRead();
```
LauraXiulan, line 35: This does not lead to a crash when trying to Read on a null stream?

src/DotNetProjectFile.Analyzers/IO/IOFile.cs
```diff
@@ -70,6 +70,22 @@ public string NameWithoutExtension
...
 
+    /// <summary>Tries <see cref="OpenRead()"/>.</summary>
+    /// <remarks>
+    /// Returns <see cref="Stream.Null"/> if reading fails.
+    /// </remarks>
+    public Stream TryOpenRead()
+    {
+        try
+        {
+            return OpenRead();
```
LauraXiulan, line 81: What is the difference between using OpenRead() and File().OpenRead()?

src/DotNetProjectFile.Analyzers/Analyzers/Generic/OnlyUseUTF8WithoutBom.cs
Corniel: The test proofs it not to fail. (That is the whole point of the `Stream.Null`, it is an empty readable stream).

src/DotNetProjectFile.Analyzers/IO/IOFile.cs
Corniel: `IOFile` is wrapper around `FileInfo` (not totally, but to a high extend). `IOFile.OpenRead()` exposes `FileInfo.OpenRead()`, by creating a `FileInfo` on the fly, and calling it's `.OpenRead()`. So method body if we would have replaced `OpenRead()` with `File().OpenRead()` we would had have the same behavior, but with code duplication.

LauraXiulan (approved):

PR #243, opened 2024-11-21
Add packages to be included as private assets part 2

Corniel:
As was a list I was still working on that was included yet when #241 was merged.


CptWesley (approved):
LGTM

PR #246, opened 2024-11-22
Proj4050: Header must be a GLOB

Corniel:
For now, only validates if the GLOB is valid to our implementation, not if is fully compatible with that of EditorConfig.

See: #201

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Rule.Ini.cs
+++ b/src/DotNetProjectFile.Analyzers/Rule.Ini.cs
@@ -1,3 +1,5 @@
+#pragma warning disable SA1118 // Parameter should not span multiple lines: readability for descriptions.
+
 namespace DotNetProjectFile;
 
 public static partial class Rule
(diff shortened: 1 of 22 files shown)


src/DotNetProjectFile.Analyzers/Rule.Ini.cs
```diff
@@ -39,5 +39,16 @@ public static class Ini
...
             category: Category.Noise);
+
+        public static DiagnosticDescriptor HeaderMustBeGlob => New(
+            id: 4050,
+            title: "Header must be a GLOB",
+            message: "Header [{0}] is not a valid GLOB.",
+            description:
+                ".editorconfig files work on the premise that header texts are " +
+                "GLOB's matching files the key-value pairs of the section apply " +
+                "to. Therefor, they must be a valid GLOB.",
```
LauraXiulan, line 50:
```suggestion
                "to. Therefor, they must be valid GLOBs.",
```

LauraXiulan (approved):
See suggestion

PR #256, opened 2024-12-23
Proj0032: Migarate away from BinaryFormatter

### LINKED ISSUE #253: New rule:  EnableUnsafeBinaryFormatterSerialization is strongly recommended against
`BinaryFormatter` has been obsolete for a [while](https://learn.microsoft.com/en-us/dotnet/core/compatibility/serialization/5.0/binaryformatter-serialization-obsolete), but still usable through `<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>`. In .NET 9 `BinaryFormatter` has been [removed](https://devblogs.microsoft.com/dotnet/binaryformatter-removed-from-dotnet-9/) and it is strongly recommended to not use it (through the unsupported `System.Runtime.Serialization.Formatters` package) and `EnableUnsafeBinaryFormatterSerialization` anymore.

Corniel:
Closes #253

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Migrate_away_from_BinaryFormatter.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Migrate_away_from_BinaryFormatter.cs
@@ -5,7 +5,7 @@ public class Reports
     [Test]
     public void when_BinaryFormatter_is_enabled() => new MigrateAwayFromBinaryFormatter()
         .ForProject("BinaryFormattingEnabled.cs")
-        .HasIssue(Issue.WRN("Proj0032", "Migrate away from BinaryFormatter and disable <EnableUnsafeBinaryFormatterSerialization>.")
+        .HasIssue(Issue.WRN("Proj0032", "Migrate away from BinaryFormatter and do not enable <EnableUnsafeBinaryFormatterSerialization>.")
             .WithSpan(04, 04, 04, 93));
 }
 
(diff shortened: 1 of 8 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Migrate_away_from_BinaryFormatter.cs
```diff
@@ -0,0 +1,19 @@
+namespace Rules.MS_Build.Migrate_away_from_BinaryFormatter;
+
+public class Reports
+{
+    [Test]
+    public void when_BinaryFormatter_is_enabled() => new MigrateAwayFromBinaryFormatter()
+        .ForProject("BinaryFormattingEnabled.cs")
+        .HasIssue(Issue.WRN("Proj0032", "Migrate away from BinaryFormatter and disable <EnableUnsafeBinaryFormatterSerialization>.")
```
CptWesley, line 8: The .csproj under tests initially enables, then later disables this flag, so one could argue it shouldn't report this options.

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -284,6 +284,16 @@ public static partial class Rule
         tags: ["clarity", "casing"],
         category: Category.Clarity);
 
+    public static DiagnosticDescriptor MigrateAwayFromBinaryFormatter => New(
+        id: 0032,
+        title: "Migrate away from BinaryFormatter",
+        message: "Migrate away from BinaryFormatter and disable <{0}>.",
```
CptWesley, line 290: I'd argue that removing the line entirely should be preferred over disabling it, but perhaps that should be its own rule

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Migrate_away_from_BinaryFormatter.cs
Corniel: There is no reason to enable it in the first place, so I'll report on setting it to true. I've even considered to report on setting it to false too due to that reasoning.

src/DotNetProjectFile.Analyzers/Rule.cs
Corniel: You're right. I've rephrased the message.

PR #262, opened 2025-01-07
Proj0033: Track uses of ""TODO"" tags

### LINKED ISSUE #250: Report on TODO/FIXME remarks
MS Build, RESX and INI files can contain TODO/FIXME phrases. We should report on those.

Corniel:
Note that there will be a follow-up also implementing this for RESX files.

Closes #250


CptWesley (approved):
LGTM

PR #267, opened 2025-01-08
Do not automatically include DotNetProjectFile.Analyzers

Corniel:
By automatically including `DotNetProjectFile.Analyzers` via a package reference, multiple issues can occur. With #259 we successfully prevented NU1504 from happening, but with `<ManagePackageVersionsCentrally>` enabled, we also have to deal with setting versions or not. It is a mess. Long story short, the SDK will not longer automatically include the analyzer, and should be added manually.

In most cases, this will not change a thing, as the the analyzer was already included for all projects via `Directory.Build.props`.

### DIFF
--- a/src/DotNetProjectFile.Analyzers.Sdk/DotNetProjectFile.Analyzers.Sdk.csproj
+++ b/src/DotNetProjectFile.Analyzers.Sdk/DotNetProjectFile.Analyzers.Sdk.csproj
@@ -28,6 +28,8 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageReleaseNotes>
       <![CDATA[
+ToBeReleased
+- Do not longer automatically include DotNetProjectFile.Analyzers
 v1.5.4
 - Prevent NU1504 from being raised when Directory.Packages.props is used. (BUG)
 v1.5.3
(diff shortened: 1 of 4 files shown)


src/DotNetProjectFile.Analyzers.Sdk/DotNetProjectFile.Analyzers.Sdk.csproj
```diff
@@ -28,6 +28,8 @@
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <PackageReleaseNotes>
       <![CDATA[
+ToBeReleased
+- Do not longer automatically include DotNetProjectFile.Analyzers
```
LauraXiulan, line 32:
```suggestion
- No longer automatically include DotNetProjectFile.Analyzers
```

LauraXiulan: Or "Do not automatically include DotNetProjectFile.Analyzers any longer"

LauraXiulan (approved):
See comments

renzo-baasdam (approved):

PR #268, opened 2025-01-08
Fix error reporting in wrong files

### LINKED ISSUE #266: ArgumentOutOfRangeException in BuildActionIncludeShouldExist
In one of my projects `BuildActionIncludeShouldExist` throws an `ArgumentOutOfRangeException `.

```text
warning AD0001: Analyzer 'DotNetProjectFile.Analyzers.MsBuild.BuildActionIncludeShouldExist' threw an exception of type 'System.ArgumentOutOfRangeException' with message 'Specified argument was out of the range of valid values.
```

CptWesley:
Closes #266 

It turned out that in many cases we assumed that the `.ReportDiagnostic` was called on the context of the file we were reporting an error in, which didn't always turn out to be a correct assumption.

I solved this by:
- hoisting the `ProjectFile` `Project` property to the `XmlAnalysisNode` interface.
- Updating `XmlAnalysisNode` implementations to match the new interface.
- Use the `ProjectFile` from the `XmlAnalysisNode` we are reporting on, rather than the one from the context.
- Added a mandatory `ProjectFile` argument to the `.ReportDiagnostic` method that takes a linespan, in order to prevent accidental incorrect usage.

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Include_package_references_once.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Include_package_references_once.cs
@@ -8,7 +8,6 @@ public class Reports
        .ForProject("DoublePackageReferences.cs")
        .HasIssues(
             new Issue("Proj0013", "Package 'Qowaiv' is already referenced.").WithSpan(04, 04, 04, 57).WithPath("DoublePackageReferences.props"),
-            new Issue("Proj0013", "Package 'Qowaiv' is already referenced.").WithSpan(04, 04, 04, 57).WithPath("DoublePackageReferences.csproj"),
             new Issue("Proj0013", "Package 'Qowaiv' is already referenced.").WithSpan(10, 04, 10, 57).WithPath("DoublePackageReferences.csproj"),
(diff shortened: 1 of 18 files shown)


CptWesley:
There are some tests failing because we assumed for some of them that the error would incorrectly be reported in the .csproj or without a path at all (oops). Will fix them.

Corniel:
Well spotted. I think I would prefer `XmlPositions` to be updated instead. I'm not sure yet, if its properties `FullSpan`, `InnerSpan`, etc. Should be come `Location`'s instead of `LinePositionSpan`'s (I think I prefer to have them as `Location`'s, but I'm not sure what the consequences could be.

CptWesley:
While I agree that having the file location somehow embedded in the positions would be a nice addition, it will be a lot of additional changes

Corniel:
> While I agree that having the file location somehow embedded in the positions would be a nice addition, it will be a lot of additional changes

What is your suggestion: to that as a follow-up? Or not at all?!

CptWesley:
> > While I agree that having the file location somehow embedded in the positions would be a nice addition, it will be a lot of additional changes
> 
> What is your suggestion: to that as a follow-up? Or not at all?!

New issue

Corniel (approved):
As mentioned, LGTM for now. The follow-up refactoring can wait for now.

PR #276, opened 2025-01-11
Proj0033: Project reference includes should exist

### LINKED ISSUE #275: New Rule: ProjectReference Should Exist
Similar to [Proj0022: Build action includes should exist](https://dotnet-project-file-analyzers.github.io/rules/Proj0022.html), but for `ProjectReference` nodes.

CptWesley:
Closes #275


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ProjectReferenceIncludeShouldExist.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    protected override void Register(ProjectFileAnalysisContext context)
+    {
+        var root = context.File.Path.Directory;
+
+        foreach (var node in context.File.Walk().OfType<ProjectReference>())
+        {
+            var include = node.Include ?? string.Empty;
+
+            if (include is not { Length: > 0 } || root.Files(include)?.Any() == false)
```
Corniel, line 17:
```suggestion
            if (include is not { Length: > 0 } || root.Files(include)?.Any() is false)
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -760,6 +760,15 @@ public static partial class Rule
         tags: ["private", "asset", "dependencies", "dependency"],
         category: Category.CodeQuality);
 
+    public static DiagnosticDescriptor ProjectReferenceIncludeShouldExist => New(
+        id: 1300,
```
Corniel, line 764: Why this high number? I would argue that this could be number 33?

PR #278, opened 2025-01-11
Fix incorrect parsing of Using node's Static property

### LINKED ISSUE #176: Using.Static should convert from nullable bool
Convert should be changed:

``` C#
public sealed class Using
{
    public bool? Static => Convert<bool?>(Attribute());
    //                                 ^
}
```

CptWesley:
Closes #176

### DIFF
--- a/src/DotNetProjectFile.Analyzers/MsBuild/Using.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/Using.cs
@@ -7,9 +7,9 @@ public sealed class Using(XElement element, Node parent, MsBuildProject project)
 
     public string? Alias => Attribute();
 
-    public bool? Static => Convert<bool>(Attribute());
+    public bool? Static => Convert<bool?>(Attribute());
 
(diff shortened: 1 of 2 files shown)


src/DotNetProjectFile.Analyzers/MsBuild/Using.cs
```diff
@@ -7,9 +7,9 @@ public sealed class Using(XElement element, Node parent, MsBuildProject project)
 
     public string? Alias => Attribute();
 
-    public bool? Static => Convert<bool>(Attribute());
+    public bool? Static => Convert<bool?>(Attribute());
 
-    public UsingType Type => (Static, Alias) switch
+    public UsingType Type => (Static ?? false, Alias) switch
```
Corniel, line 12: Maybe define the last one as `_ =>` and put in comment that it is `(true, _)`?

Corniel (approved):
Besides remark, LGTM

Corniel (approved):
LGTM!

PR #290, opened 2025-01-14
Proj1011 report on PackageVersion nodes if applicable

### LINKED ISSUE #288: Proj1011 Should probably report on the PackageVersion node rather than the PackageReference node
Currently [Proj1011](https://dotnet-project-file-analyzers.github.io/rules/Proj1101.html) always reports on the `PackageReference` node, even when the unstable version is actually specified by a seperate `PackageVersion` node

CptWesley:
Closes #288


Corniel (approved):
LGTM!

PR #293, opened 2025-01-15
Fix TUnit related false positives

CptWesley:
- TUnit.Analyzer and TUnit.Assertions.Analyzer are deprecated, so they shouldn't be recommended anymore (the analyzers are now included in the main packages).
- TUnit doesn't use VSTest and therefore doesn't require the test SDK (it even breaks if the test SDK is loaded).


Corniel (approved):
LGTM!

PR #298, opened 2025-01-22
INI grammar

Corniel:
Replaces the home grown INI parser with an ANTLR4 one.


CptWesley:
Do you have some benchmarks for this? I would be curious what the difference is

Corniel:
> Do you have some benchmarks for this? I would be curious what the difference is

| Method       | Index | Mean         |
|------------- |------ |-------------:|
| Home Grown |  27 LoC     |     126.0 us |
| ANTLR4  | 27 LoC    |     185.6 us |
| Home Grown | 36 LoC     |     140.2 us |
| ANTLR4  | 36 LoC     |     193.7 us |
| Home Grown | 1220 LoC    | 225,944.9 us |
| ANTLR4  | 1220 LoC    |   1,880.8 us |

It scales pretty well.

Corniel:
@CptWesley Locally, it builds like a charm, but It seems that the code is not generated on the build, do you have any clue?

CptWesley:
> @CptWesley Locally, it builds like a charm, but It seems that the code is not generated on the build, do you have any clue?

@Corniel Probably the same reason why it doesn't build for me locally either 😜

CptWesley:
@Corniel First attempt at building doesn't yet include the generated sources, second time around it does, perhaps things are not guaranteed to build in the right order? I'm not sure how the generation works (it's not a modern Roslyn source generator sadly).

CptWesley:
@Corniel It's the build order. If `DotNetProjectFile.Analyzers` tries to build before `Antlr4.Grammars`, things break.

CptWesley:
@Corniel Added explicit build order dependency between the projects to the `.sln` file

Corniel:
@CptWesley Apparently, but enforcing the build order did not help, it seems.

CptWesley:
> @CptWesley Apparently, but enforcing the build order did not help, it seems.

Yeah, I noticed... At least it fixed the issue of it not building locally after a clean.

CptWesley:
It appears that the `Compile Include` does not find the files for some reason

Corniel:
> It appears that the `Compile Include` does not find the files for some reason, because referencing the types directly from `Antlr4.Grammars` works fine

I considered to copy them instead or to add the generator to the target project. The latter is also not ideal, as the generator does not work with the included files, but still needs a package dependency (that could then be private). But also from an project structure perspective I prefer the have a separate project only containing the Grammar files I think.

CptWesley:
@Corniel I think it should work now...

src/DotNetProjectFile.Analyzers/EditorConfig/EditorConfigFile.cs
```diff
@@ -9,19 +9,20 @@ public sealed class EditorConfigFile(IniFileSyntax syntax) : ProjectFile
...
     public WarningPragmas WarningPragmas => WarningPragmas.None;
 
     public bool IsRoot
         => Syntax.Sections.FirstOrDefault() is { } section
         && section.Header is null
-        && section.Kvps
-            .Where(kvp => kvp.Key.IsMatch("root"))
-            .Select(kvp => kvp.Value.IsMatch("true"))
-            .LastOrDefault();
+        //&& section.Kvps
```
CptWesley, line 23: intentionally commented?

src/DotNetProjectFile.Analyzers/EditorConfig/EditorConfigFile.cs
```diff
@@ -9,19 +9,20 @@ public sealed class EditorConfigFile(IniFileSyntax syntax) : ProjectFile
     public IniFileSyntax Syntax { get; } = syntax;
 
     /// <inheritdoc />
-    public IOFile Path => Syntax.SyntaxTree.Path;
+    public IOFile Path => default; // Syntax.SyntaxTree.Path;
```
CptWesley, line 12: intentional?

PR #305, opened 2025-02-07
Report on RESX files that are not added as additional files

Corniel:
I was under the impression for some time that embedded resources where added as additional files by default, and that the only reason our test tools did pick add them to the additional files was due to an issue with those tools (Buildalyzer). I apparently never tested this assumption correctly, or the behavior did change (way less likely).

Long story short, with this fix, all resource files that are not added as additional files are also analyzed.


CptWesley (approved):
LGTM

PR #306, opened 2025-02-07
Proj0006: Report on RESX files that are not added as additional files

Corniel:
See #305 . So Proj0006 should be extended. This also means that the documentation should be extended. Not only that, but the rule now becomes a generic rule (applies both to RESX, and MS Build files), we could to introduce a new rule ID for it (and mark the old one as obsolete, of drop it).


CptWesley (approved):

CptWesley:
I think the ID can stay the same, since the 0006 is in the range of things that need to be added to the csproj/props files, which is still the case here, even if it is a .resx file.

Issue #314, opened 2025-02-13
Proj0809 has no knownlegde on non-private asset packages


Corniel:
The implementation of Proj0809 relies on knowing a package is not meant to be a private asset. However, in the list with known packages, currently only packages that should be a private asset are included (analyzers, code generators, SDK's). Therefor it will for now never report on anything. It would be useful to consider a way to detect if a package should be considered a private asset or not in a different way.

CptWesley:
Does our analyzer have the magic power to find the correct .nupkg and inspect the contents? If there is only a .dll in the analyzer folder and not in the runtime folder it is quite safe to assume it's not meant to do anything at runtime.

Corniel:
For [Roslyn Test Tools ](https://github.com/Corniel/roslyn-test-tools) I download (with some caching) the NuGet package (with some caching). I'm not aware of another way to do so. If we know and have access to the path where the NuGet packages are cached we might be able to do so. That would also give is other possibilities (check for licenses etc,). I'm not aware how.

Your suggested approach might be doable fallback.

CptWesley:
The path of the nuget cache is quite deterministic: https://learn.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders

tl;dr:
- Path of environment variable `NUGET_PACKAGES` if available
- `globalPackagesFolder` in nuget config if available
- `repositoryPath` in `packages.config` if available
- `%userprofile%\.nuget\packages` if Windows
- `~/.nuget/packages` if Unix

Corniel:
@CptWesley Using (a part of) the `NuGet.Client` package might be the best solution. My point was more that the logic involved so resolve and read the nuspec files is substantial.

CptWesley:
For this particular issue it would be sufficient to: find the corresponding directory that matches the package, picking the right (or close-enough) version number and inspecting the contents.

- Find cache directory (see my previous comment)
- Find directory that matches package (it appears a case insensitive match would be sufficient)
- Find directory matching the package version (or a close-enough version if not found, realistically it is unlikely that this info changes much for different versions anyway).
- Check if directory contains a `.dll` in the `analyzer` and/or `lib` sub-directories.

PR #315, opened 2025-02-15
Proj0501: Rely on compliant licenses only

### LINKED ISSUE #322: Rely on compliant licenses only
Now we're able to [resolve .nuspec files](#316), we also can build rules involving which licenses to support. In fact with #315 (now in draft) the first steps in that direction have been made.

The most important question however is on how to configure which licenses are compliant.

From a configuration perspective, the seems ideal:

``` XML
<project>
  <ProperyGroup>
    <AllowedLicenses>MIT;Apache2.0</AllowedLicenses>
    <AllowedPackages>MyCompany.*</AllowedPackages>
  </ProperyGroup>
</project>
```

Another thing would be that I think that `MIT` and `Apache2.0` (it seems that Microsoft uses the latter for most/all of its packages) should be included if nothing is specified. Should there be others that should be included automatically?

I would argue that packages not containing a package should always be added in the list of `<AllowedPackages>`. Using packages without an explicit license is - imho - a bad practice.

But getting the right value (taking conditionals and such into account) is something we do not have fully covered. That being said, it might be worth checking: See: https://platform.uno/blog/using-msbuild-items-and-properties-in-c-9-source-generators/

Corniel:
As a result of #314 , I thought should also be possible to check for the used licenses. We have to tackle the dependency with NuGet, but once we have, we can do a lot of nice things.

Closes #322

### DIFF
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -44,6 +44,8 @@
 
   <ItemGroup Label="Roslyn dependencies">
     <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.10.0" ExcludeAssets="runtime" />
+    <PackageReference Include="NuGet.Configuration" Version="6.13.1" />
+    <PackageReference Include="NuGet.Packaging" Version="6.13.1" />
   </ItemGroup>
 
   <ItemGroup Label="Analyzers">
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Rely_on_compliant_licenses_only.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Rely_on_compliant_licenses_only.cs
@@ -1,0 +1,29 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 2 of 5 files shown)


src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -44,6 +44,8 @@
 
   <ItemGroup Label="Roslyn dependencies">
     <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.10.0" ExcludeAssets="runtime" />
+    <PackageReference Include="NuGet.Configuration" Version="6.13.1" />
+    <PackageReference Include="NuGet.Packaging" Version="6.13.1" />
```
CptWesley, line 48: I don't think we really need this dependency to be honest, it is quite straightforward to find the .nuspec file and in there we honestly only care about 1 or 2 fields `package` -> `metadata` -> `license` and/or `package` -> `metadata` -> `licenseUrl`

Corniel:
Continue with this after #316 has been merged.

CptWesley:
@Corniel I have a local branch that adds parsed license expressions (with some logic for defining and checking compatbility) obtained from the `.nuspec` files to the `CachedPackage` object. I want to open a draft PR for that once #316 is merged so that this rule (`Proj0500`) can just use that logic.

Corniel:
Abandoned this one in favor of #336

PR #317, opened 2025-02-15
Implement the NuSpec file

Corniel:
To be able to analyze `.nuspec` files it is beneficial to have a well defined representation of its structure.


CptWesley:
I'm just wondering if it really matters to have this, considering we only care about the license expression field, I don't think we will ever care about the author of a package that is being referenced...

Corniel:
> I'm just wondering if it really matters to have this, considering we only care about the license expression field, I don't think we will ever care about the author of a package that is being referenced...

Not really, but according to the documentation, those fields are mandatory,  but I can drop them. ;)

CptWesley:
@Corniel since these are from packages that have already been pushed to nuget and not in control of the person running the analyzer, it is not really our concern to validate the correctness

CptWesley (approved):

PR #325, opened 2025-02-16
Use cache for dependency types

Corniel:
Rule [Proj1200: Exclude private assets as project file dependency](https://dotnet-project-file-analyzers.github.io/rules/Proj1200.html) currently has a fixed set of packages we advise it for. With the cache in place, and also collecting the `<developmentDependency>` I think it would be better to start using that data to give advise when to markreferences as private assets.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ExcludePrivateAssetDependencies.cs
+++ b/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ExcludePrivateAssetDependencies.cs
@@ -2,6 +2,7 @@ using DotNetProjectFile.NuGet;
 
 namespace DotNetProjectFile.Analyzers.MsBuild;
 
+/// <summary>Implements <see cref="Rule.ExcludePrivateAssetDependencies"/>.</summary>
 [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
 public sealed class ExcludePrivateAssetDependencies() : MsBuildProjectFileAnalyzer(Rule.ExcludePrivateAssetDependencies)
 {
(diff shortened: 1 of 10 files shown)


src/DotNetProjectFile.Analyzers/NuGet/PackageCache.cs
```diff
@@ -174,4 +176,18 @@ bool HasDllFiles(string subDir)
             return null;
         }
     }
+
+    private static NuSpecFile? TryLoadNuSpecFile(IODirectory directory)
+    {
+        if(directory.Files("*.nuspec")?.FirstOrDefault() is { HasValue: true} nuspec)
```
CptWesley, line 182: I think it's better to try to see if there is a nuspec file with the exact lower-case package name (in case for some reason someone included a second nuspec file) before grabbing a random one. Also I don't think `HasValue: true` is necessary.

Corniel: I agree. I've update it.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ExcludePrivateAssetDependencies.cs
```diff
@@ -29,14 +30,9 @@ private static bool ShoudBePrivateAssets(PackageReference reference)
             return false;
         }
 
-        if (NuGet.Packages.All.TryGet(reference.Include) is { } pkg)
```
CptWesley, line 32: You want to drop the manually maintained list for this analyzer?

CptWesley (approved):

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ExcludePrivateAssetDependencies.cs
Corniel: I do not think its add any value now we can access the nuspec files. We need those for licenses anyway, so they are one way or the other needed and cached for later usage. We could consider adding some extra data to it, that we want to define (like analyzers being language specific and or general purpose)

Issue #330, opened 2025-02-17
Skip rules that rely on imports being resolved once that fails


Corniel:
Once an import statement could not be resolved, rules relying on imports (most rules) should not be performed as they will likely report FP's.

Issue #321, opened 2025-02-16
AD0001 warnings for test projects


mmanovic:
Here [https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022#use-case-multi-level-merging](url) it can be seen common solution structure where test projects are located within test subfolder that contains Directory.Build.props file. 

This code is a common way to inherit/merge settings from the base Directory.Build.props file located at the solution file level.
`<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
`

That approach throws many AD0001 warnings for all test projects within test folder. Here is an example of warning:
 CSC : warning AD0001: Analyzer 'DotNetProjectFile.Analyzers.MsBuild.UseVersionOverrideOnlyWithCpm' threw an exception of type 'System.NotSupportedException' with message 'The given path's format is not supported.'.

The warning message leads to the conclusion that there is some problem within DotNetProjectFile.Analyzers when it comes to path parsing/resolving on Windows machines. The current workaround that is working for me is to reference base Directory.Build.props file directly with hardcoded relative path as it can be seen on the following screenshot:
`  <Import Project="../Directory.Build.props" />
`
![Image](https://github.com/user-attachments/assets/90684ec6-9bf6-4e2a-a666-32dce9f3f663)

Corniel:
Thanks for reporting this issue. An `AD0001` is always bad: it meant an analyzer crashed while running. We do not try to execute expressions in paths (or conditionals); while running an analyzer, Roslyn does not provide enough context to do so.

I would solve this differently: have a `Directory.Build.props` that only contains settings that are shared by all projects, and define a `test.props` (or other .props file for a group of projects) that contains settings only for test (or other) projects.

CptWesley:
@Corniel IMO It would still be better to skip the rule rather than to spam warnings because the analyzer doesn't support it

Corniel:
@CptWesley I'm not aware that I suggested otherwise. My suggestion was on how to implement this, not because we do not support this, but because I think it is more readable. If we do not support/understand something, we should definitely not crash, and Ideally not report false positives.

mmanovic:
> [@Corniel](https://github.com/Corniel) IMO It would still be better to skip the rule rather than to spam warnings because the analyzer doesn't support it

[...]

PR #331, opened 2025-02-17
Proj0034: Import statement could not be resolved

Corniel:
As a result of #321, we should inform the use that we could not resolve the import, and do not perform analysis on rules that require the import statements to work (See #330 ).

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
@@ -1,0 +1,20 @@
+using DotNetProjectFile.MsBuild;
+
+namespace Rules.MS_Build.Unresolvable_Import;
+
+public class Reports
+{
+    [Test]
+    public void empty_includes() => new UnresolvableImport()
+    .ForProject("UnresolvableImport.cs")
+        .HasIssue(Issue.WRN("Proj0034", "The <Import> '$(MSBuildThisFileDirectory)common.props' could not be resolved.").WithSpan(02, 02, 02, 62));
+}
+
+public class Guards
+{
+    [TestCase("CompliantCSharp.cs")]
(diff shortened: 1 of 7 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
```diff
@@ -0,0 +1,20 @@
+using DotNetProjectFile.MsBuild;
+
+namespace Rules.MS_Build.Unresolvable_Import;
+
+public class Reports
+{
+    [Test]
+    public void empty_includes() => new UnresolvableImport()
+    .ForProject("UnresolvableImport.cs")
+        .HasIssue(Issue.WRN("Proj0034", "The <Import> '$(MSBuildThisFileDirectory)common.props' could not be resolved.").WithSpan(02, 02, 02, 62));
```
CptWesley, line 10: Perhaps this warning message gives the false impression that the import is incorrectly defined, rather than the analyzers simply not knowing how to handle it.

Corniel: Suggestion how to improve on that?

Corniel: I added "... by the analyzer."

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -60,6 +60,7 @@
     <PackageReleaseNotes>
       <![CDATA[
 ToBeReleased:
+- Proj0034: Import statement could not be resolved. (NEW RULE)
```
CptWesley, line 63: Update message here as well

CptWesley (approved):

PR #332, opened 2025-02-17
Disable rules that can not be excecuted correctly on a failing import

### LINKED ISSUE #330: Skip rules that rely on imports being resolved once that fails

Corniel:
When an import fails (due to the fact that the analyzer can resolve the path (reported by [Proj0034](https://dotnet-project-file-analyzers.github.io/rules/Proj0034.html)) a big chunk of the rules can not be correctly executed, as the outcome of the analysis is likely effected by what is defined in the (failed) import.

All MS Build file rules are now disabled in that case, unless specified otherwise.

Closes #330 and #321

### DIFF
--- a/projects/UnresolvableImport/UnresolvableImport.csproj
+++ b/projects/UnresolvableImport/UnresolvableImport.csproj
@@ -6,6 +6,8 @@
     <TargetFramework>net8.0</TargetFramework>
   </PropertyGroup>
 
+  <!-- TODO: should be reported on./ -->
+  
   <ItemGroup>
     <Compile Include="../common/Code.cs" />
   </ItemGroup>
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
@@ -1,17 +1,20 @@
-using DotNetProjectFile.MsBuild;
-
 namespace Rules.MS_Build.Unresolvable_Import;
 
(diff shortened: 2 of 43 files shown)


projects/UnresolvableImport/UnresolvableImport.csproj
```diff
@@ -6,6 +6,8 @@
     <TargetFramework>net8.0</TargetFramework>
   </PropertyGroup>
 
+  <!-- TODO: should be reported on./ -->
```
CptWesley, line 9: Is this intentionally left here?

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
```diff
@@ -1,17 +1,20 @@
-using DotNetProjectFile.MsBuild;
-
 namespace Rules.MS_Build.Unresolvable_Import;
 
 public class Reports
 {
     [Test]
-    public void empty_includes() => new UnresolvableImport()
```
CptWesley, line 8: why was this test case removed?

CptWesley (approved):

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Unresolvable_Import.cs
Corniel: That was by mistake, reverted.

PR #333, opened 2025-02-17
Introduction of `GetMsBuildProperty()`

Corniel:
The execution (and reading) of actual values it a huge improvement to be able to rely on the actual values of MS Build, instead of walking through the full XML tree. For most values, we have to add those using `<CompilerVisibleProperty>`. By adding a props file to the package, users are not longer enforced to do that themselves.


CptWesley:
Potential problem here is that this returns the final value, rather than the value at the point where the node is defined/interpreted

Corniel:
> Potential problem here is that this returns the final value, rather than the value at the point where the node is defined/interpreted

That surely is a challenge, and we should be aware of it, but I think this could be a good way to resolve at least some of the values.

CptWesley (approved):

PR #335, opened 2025-02-18
Proj0500: Only include packages with an explicitly defined license

Corniel:
Before reporting on not allowed licenses, I thought we first might want to report on packages without  (explicitly defined) licenses.

Note that the test that resports is to a high extend a FP: that Nuspec contains a `<licenseUrl>` instead. I hope that we actually can find a package that is beeing shipped without, so that we can have a true valid issue being reported in the test.

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Only_include_packages_with_explicit_license.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Only_include_packages_with_explicit_license.cs
@@ -1,0 +1,29 @@
+namespace Rules.MS_Build.Only_include_packages_with_explicit_license;
+
+public class Reports
+{
+    [Test]
+    public void on_packages_without_license_specified_in_nuspec() => new OnlyIncludePackagesWithExplicitLicense()
+       .ForInlineCsproj(@"
+<Project Sdk=""Microsoft.NET.Sdk"">
+
+    <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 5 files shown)


CptWesley:
If a `licenseUrl` is present then some form of license is still provided. So I'm not sure if you want to report on that. Besides that, similar issues occur when projects use license files rather than license expressions (be it permissive licenses or custom licenses). Furthermore, even if a dependency is copy righted, that doesn't mean you can't use it. It is possible that you have a licensing agreement with the providing party. So I'm not 100% if this is desireable without causing a large number of FPs right now

Corniel:
@CptWesley as stated, we should exclude `.nuspec` files with a `<licenseUrl>'. For now, it proves that the rule works. How to allow packages that have a license file instead of an expression is another issue. I've an idea how to do that, but let discuss that there.

CptWesley:
@Corniel `Microsoft.DotNet.PlatformAbstractions` version `1.1.1` lacks both `license` and `licenseUrl`

Corniel:
@CptWesley I updated the test case to use `Microsoft.DotNet.PlatformAbstractions`, and added a test for a compliant LicenseUrl only.

projects/WithRuleset/WithRuleset.csproj
```diff
@@ -7,6 +7,7 @@
   
   <ItemGroup>
     <Compile Include="../common/Code.cs" />
+    <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="1.1.1" />
```
CptWesley, line 10: Why added here? This file does not seem to be referenced by any of the test cases

CptWesley (approved):

CptWesley (approved):

PR #336, opened 2025-02-18
Proj0502: Package contains an incompatible license

### LINKED ISSUE #322: Rely on compliant licenses only

Corniel:
A clean start for Proj0502: Only include packages with a compliant license. After finishing  #335 ,  #315 became hard to forward integrate, also due to some other rather useful changes that have been made in the meantime.

Closes #322


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
```diff
@@ -7,32 +6,59 @@ namespace DotNetProjectFile.Analyzers.MsBuild;
...
+    Rule.PackageContainsIncompatibleLicense)
 {
     /// <inheritdoc />
     public override bool DisableOnFailingImport => false;
 
     /// <inheritdoc />
-    protected override void Register(ProjectFileAnalysisContext<MsBuildProject> context)
+    protected override void Register(ProjectFileAnalysisContext context)
     {
+        var allowed = (context.GetMsBuildProperty("AllowedLicenses") ?? string.Empty)
```
CptWesley, line 18: Shouldn't this default to `Licenses.All` if no explicit allow list has been provided? That way it just assumes anything is fine as long as the licenses don't violate each other.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
```diff
@@ -7,32 +6,59 @@ namespace DotNetProjectFile.Analyzers.MsBuild;
...
+            return;
+        }
+
+        var expression = package.LicenseExpression();
+
+        if (package.UrlOnly() && expression.IsUnknown)
+        {
+            context.ReportDiagnostic(Rule.PackageOnlyContainsDeprecatedLicenseUrl, reference, reference.IncludeOrUpdate);
+        }
+        else if (expression.IsKnown && allowed.None(l => l.CompatibleWith(expression)))
```
CptWesley, line 45: This is checking that at least one allowed license could be used a dependency for the dependency that is being analyzed. I don't think that is what you want to check here.

I expect you would want to either check that `!allowed.Contains(expression) || !expression.CompatibleWith(<ParsedLicense-node-of-this-project>)`

CptWesley: `a.CompatibleWith(b)` checks that a project with license `b` is allowed to have a package with license `a` as a dependency.

CptWesley: For example:

```csharp
GPL3.CompatibleWith(AGPL3) // true
AGPL3.CompatibleWith(GPL3) // false

GPL3.CompatibleWith(MIT) // false
MIT.CompatibleWith(GPL3) // true

BSD3.CompatibleWith(MIT) // true
MIT.CompatibleWith(BSD3) // true

GPL3.CompatibleWith(GPL2) // false
GPL2.CompatibleWith(GPL3) // false
```

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
Corniel: I think we should enforce this, and it would be fine to (if not specified) allow all permissive licenses (we might even want to have a single expression that contains all together) if not specified otherwise, but reject copy-left licenses unless explicitly allowed.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
Corniel: The second part of the expression is a really nice idea! It should have a (slightly) different message I think. It should be explicit that the reference is incompatible with the license of the project's package it self, rather than being not being allowed because it is not in some fixed list.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
CptWesley: But if you're own license is AGPL3 or GPL3 there is not really much point in requiring people to manually add GPL3 to their allow-list in my opinion, since the compatibility is already known. I would expect manually setting the list of allowed licenses is mostly useful for companies that only allow a certain pre-verified list of licenses to be used

Corniel: That is fair. As most projects are not packages their own, I suggest to only implement that scenario for now, and in a follow-up take the compatibility of the project itself into account. The more I think of it, the more I think the logic and communication to the user benefits from having a different rule stating the the package has a license that conflicts the license of the project.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
```diff
@@ -7,32 +6,62 @@ namespace DotNetProjectFile.Analyzers.MsBuild;
...
+        {
+            context.ReportDiagnostic(Rule.PackageContainsIncompatibleLicense, reference, reference.IncludeOrUpdate, expression);
+        }
+    }
+
+    private static ImmutableArray<LicenseExpression> AllowedLicenses(ProjectFileAnalysisContext context)
+        => context.GetMsBuildProperty("AllowedLicenses") is { } custom
+        ? [..custom.Split([',', ';'], StringSplitOptions.RemoveEmptyEntries)
+            .Select(l => Licenses.FromExpression(l.Trim()))]
+        : Licenses.Permissive;
```
CptWesley, line 52: I still don't get why not just use `Licenses.All` here, since the other rule will cover that issue, since the copy-left licenses are not compatible with `Unknown`

CptWesley: Again the same point I made earlier, if your application is licensed under e.g. GPL3, why would you then still be required to add GPL3 to your allow list to not trigger a warning here?

Currently you run into these warnings when using `Licenses.Permissive` as a default:
```
GPL3 with GPL3 dependency: Proj0502 (strange in my opinion)
GPL3 with MIT dependency: no warning
MIT with GPL3 dependency: Proj050x incompatible license and Proj0502 (redundant)
Unknown with GPL3 dependency: Proj050x incompatible license and Proj0502 (redundant)
```

If you were to default to `Licenses.All` these scenarios would occur when not defining an allow list:
```
GPL3 with GPL3 dependency: no warnings
GPL3 with MIT dependency: no warning
MIT with GPL3 dependency: Proj050x incompatible license
Unknown with GPL3 dependency: Proj050x incompatible license
```

CptWesley: Therefore I think it is better to default to allowing everything and let the compatibility rule catch the transgressions

Corniel: See: https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers.github.io/pull/93

I think you're right. We should focus on licenses that are compatable with the license of the project first (and this rule should be about that). Than, if we could not resolve the license of the project, or the package, we should have other rules dealing with that.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
```diff
@@ -7,32 +6,56 @@ namespace DotNetProjectFile.Analyzers.MsBuild;
...
+            return;
+        }
+
+        var packageLicense = package.LicenseExpression();
+
+        if (package.UrlOnly() && packageLicense.IsUnknown)
+        {
+            context.ReportDiagnostic(Rule.PackageOnlyContainsDeprecatedLicenseUrl, reference, reference.IncludeOrUpdate);
+        }
+        else if (projectLicense.IsKnown && !packageLicense.CompatibleWith(projectLicense))
```
CptWesley, line 42: We shouldn't need this `projectLicense.IsKnown` here, I would argue it should be the responsibility of the different licenses to determine if they can be used in any case in their `CompatibleWith` implementation.

Corniel: That is true once we have a good way to deal with:
- LicenseExpression None is supported
- CustomLicense (LicenseFile is specified which should be compatible then)

I would like to do that in follow-up PR's.

CptWesley: I fail to see how that is relevant. Permissive ones will always return `true` and the strict copyleft ones will return `false`, when the license is `None`, `Custom` or `Unknown`, so that's not different from the current behaviour of `Unknown`

Corniel: Well, with `None` it should be true for all, as you do not share anything. But we can say that we will not ship in the meantime, so that is not an issue.

CptWesley: I would argue against that, because if a dependency is GPL3 and your own license is None, you are in violation of GPL3

PR #339, opened 2025-02-18
License URL parsing

CptWesley:
Parse licenses expressions.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Licensing/Licenses.cs
+++ b/src/DotNetProjectFile.Analyzers/Licensing/Licenses.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Frozen;
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


src/DotNetProjectFile.Analyzers/Licensing/Licenses.cs
```diff
@@ -93,7 +107,18 @@ public static LicenseExpression FromExpression(string? licenseExpression)
 
     public static LicenseExpression FromUrl(string? licenseUrl)
```
Corniel, line 108: It makes more sense to define this on LicenseExpression itself I guess.

CptWesley: Not so sure about that. If we want to refactor it, I would like to do it in a different PR.

PR #341, opened 2025-02-18
Generator for extracting SPDX License data

CptWesley:
Scrapes SPDX license data.

- License texts are downloaded once by the generator and then stored in the repository (to save time requesting the same file every build)
- SPDX info is binary encoded and compressed to ensure the .nupkg file does not grow to massive sizes when support for more licenses is added.


Corniel:
I Think I prefer to have the scraped licenses as embedded text.

Corniel:
Now that directory contains txt files, it might be worth to rename it to `Sources`.

Corniel (approved):
LGTM!

PR #343, opened 2025-02-19
Proj0217: Define requiring license acceptance explicitly

### LINKED ISSUE #342: New Rule: `PackageRequireLicenseAcceptance` should be explicitly defined
Setting `PackageRequireLicenseAcceptance` to `true` in the `.csproj` causes the `RequireLicenseAcceptance` flag to be enabled in the `.nuspec` which will result in consumers of the package being prompted about the library package when installed through software that supports this feature (e.g. Visual Studio's NuGet package manager).

CptWesley:
Closes #342

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 ![.NET project file analyzers logo](design/logo_128x128.png)
 # .NET project file analyzers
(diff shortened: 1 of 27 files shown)


Corniel (approved):
Please don't forget to update the release notes too.

PR #349, opened 2025-02-20
Determine license from nuget package license files

### LINKED ISSUE #350: Determine license from license files
`.nuspec` files can also define a license based on a license file, rather than an expression or a URL. We can make a good estimate of what the license is by trying to match it to known licenses.

CptWesley:
Closes #350

Determines package license based on the contents of provided license files.

Algorithm checks if file similarity (after stripping punctuation and whitespace) is 95% based on the Sørensen–Dice coefficient with bigrams. This algorithm is similar to the one used by https://github.com/licensee/licensee (the tool used by GitHub to fuzzily determine the license).


src/DotNetProjectFile.Analyzers/TextSimilarity/JaccardIndex.cs
```diff
@@ -0,0 +1,29 @@
+namespace DotNetProjectFile.TextSimilarity;
```
Corniel, line 1: Why not just Text? Would do for me.

Corniel (approved):
Besides my nitpicking: LGTM!

PR #352, opened 2025-02-20
Support more (common) license types

CptWesley:
Added SPDX licenses:
- `EPL-1.0`
- `EPL-2.0`
- `MS-LPL`
- `PostgreSQL`
- `WTFPL`

Added non-SPDX licenses:
- `NET_Library_EULA_ENU` this is the old distributable license that Microsoft would use for distributing their binary files before switching to `MIT` or `Apache-2.0` in more recent times. Older targets and some less popular libraries use this.

The latter addition also resulted in me adding a mechanism to inject non-spdx licenses into the spdx-license-info generator.

Locally I am now able to identify the license of `4406` out of `4906` (roughly `90%`) packages in my nuget cache 🎉.


Corniel (approved):
LGTM!

What is the reason for moving the generated (downloaded/resolved) License texts?

CptWesley:
@Corniel mostly so you don't accidentally delete the manually added ones if you delete the entire folder to force them to be redownloaded (which I already did twice when working on this PR)

PR #353, opened 2025-02-20
Proj0035: Remove deprecated RestoreProjectStyle property

### LINKED ISSUE #236: New Rule: Suggest to use <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
According to [Microsoft](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#using-packagereference-for-a-project-with-no-package-dependencies) this can be useful, but I'm not certain if I understand the usage correctly: 

> Advanced: If you have no packages installed in a project (no PackageReferences in project file and no packages.config file), but want the project to be restored as PackageReference style, you can set a Project property RestoreProjectStyle to PackageReference in your project file.
> ``` XML
>
><PropertyGroup>
>    
>    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
>    
></PropertyGroup>    
> ```
> This may be useful, if you reference projects which are PackageReference styled (existing csproj or SDK-style projects). This will enable packages that those projects refer to, to be "transitively" referenced by your project.

Is it *always* advisable for projects with `<ProjectReference>`'s, but without  `<PackageReference>`s (this is easy to detect), and why?! Having a rule that says: MS is saying it might help, but they did not tell why, and we have no clue, does not make sense.

Corniel:
Closes #236

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Remove_deprecated_estoreProjectStyle.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Remove_deprecated_estoreProjectStyle.cs
@@ -1,0 +1,27 @@
+namespace Rules.MS_Build.Remove_deprecated_estoreProjectStyle;
+
+public class Reports
+{
+    [Test]
+    public void when_property_is_defined() => new RemoveDeprecatedRestoreProjectStyle()
+       .ForInlineCsproj(@"<Project Sdk=""Microsoft.NET.Sdk"">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
+  </PropertyGroup>
+
+</Project>")
+       .HasIssue(Issue.WRN("Proj0035", "Remove the <RestoreProjectStyle> property")
+       .WithSpan(04, 04, 04, 63));
+}
+
+public class Guards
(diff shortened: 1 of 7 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Remove_deprecated_estoreProjectStyle.cs
```diff
@@ -0,0 +1,27 @@
+namespace Rules.MS_Build.Remove_deprecated_estoreProjectStyle;
```
LauraXiulan, line 1: Fix filename and namespace

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -311,6 +311,17 @@ public static partial class Rule
...
         category: Category.Bug);
 
+    public static DiagnosticDescriptor RemoveDeprecatedRestoreProjectStyle => New(
+        id: 0035,
+        title: "Remove deprecated RestoreProjectStyle property",
+        message: "Remove the <RestoreProjectStyle> property",
+        description:
+            "The <RestoreProjectStyle> was once introduced to mitigate third-party " +
+            "references using a packages.config file. This practice has been " +
+            "depericated for a very long time now.",
```
LauraXiulan, line 321:
```suggestion
            "deprecated for a very long time now.",
```

PR #362, opened 2025-02-21
Introduction of custom license

Corniel:
To make a clear distinction between unknown and custom licenses (such as that of Sonar Source, or XCeed) I've created a custom license, containing an (SHA1) hash.

The idea is that custom licenses should be registered, to be allowed. Previously I've suggested to define this in a MS Build property, but an MS Build item might be more convenient:

``` XML
<ItemGroup Label="Custom licenses">
  <CustomLicense Include="FluentAssertions" Hash="RQN3OPC/wYxT+eMudU1hVTy7smo" />
</ItemGroup>
```

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Thrid_Party_license_compliance.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Thrid_Party_license_compliance.cs
@@ -54,6 +54,7 @@ public class Reports
 </Project>")
        .HasIssue(Issue.WRN("Proj0502", "The SeeSharpTools.JY.GUI package is distributed as GPL-3.0-only, which is imcompatable with the MIT license of the project."));
 
+    [Ignore("Rule 0504 has not been implemented yet")]
     [Test]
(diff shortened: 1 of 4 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Thrid_Party_license_compliance.cs
```diff
@@ -53,6 +53,23 @@ public class Reports
...
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include=""FluentAssertions"" Version=""8.0.0"" />
+  </ItemGroup>
+
+</Project>")
+       .HasIssue(Issue.WRN("Proj0504", "FluentAssertions package is distributed with a custom license, which has not been registrated."));
```
CptWesley, line 72:
```suggestion
       .HasIssue(Issue.WRN("Proj0504", "FluentAssertions package is distributed under a custom license, which has not been explicitly accepted."));
```

CptWesley (approved):
Looks good so far otherwise

PR #363, opened 2025-02-21
Resolve nuget.config globalPackagesFolder

### LINKED ISSUE #319: `PackageCache` Handle `nuget.config`
Currently the `PackageCache` is only capable of handling the default cache directories or the one set by the `NUGET_PACKAGES` environment variable, but it should also be able to handle `globalPackagesFolder` set by the `nuget.config`

See also: #314 and #318

Corniel:
Closes #319

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/NuGet/NuGet_config_file_specs.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/NuGet/NuGet_config_file_specs.cs
@@ -1,0 +1,48 @@
+using DotNetProjectFile.NuGet.Packaging;
+using Specs.TestTools;
+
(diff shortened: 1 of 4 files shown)


CptWesley:
Maybe eventually yes, but for now I think we can just use the content without applying any rules to it.

Corniel:
I'm not sure what the best way is to continue now. From the directory of the project file, potentially up to root, a check should be done if there is `nuget.config` file. This potentially implies that the cache should give a different result for different projects: the custom location could theoretically resolve a different nuspec file for the same PackageId Version combination. And where the path is set now, we do not have the scope of hte project file directory yet.

Corniel:
We can already merge it but not use it yet.

CptWesley (approved):

PR #364, opened 2025-02-21
Add support for multiple license texts

### LINKED ISSUE #358: Licensing: Support alternative license texts
Some licenses are used with varying license texts. Most notably the `Apache-2.0` license which has a short and a long version, where we currently only check for the short version. Microsoft also appears to be using a second variation of the `NET_Library_EULA_ENU` which specifically mentions `.js` when reffering to the distributed work for their web related libraries.

CptWesley:
Closes #358

Also added a few more permissive licenses that I encountered in my nuget cache


src/DotNetProjectFile.Analyzers/Licensing/SpdxLicenseInfo.cs
```diff
@@ -66,7 +64,7 @@ public static SpdxLicenseInfo ReadFrom(BinaryReader reader)
             Osi = osi,
             Fsf = fsf,
             SeeAlso = seeAlso.ToImmutableArray(),
-            LicenseText = licenseText,
+            LicenseTexts = licenseTexts.ToImmutableArray(),
```
Corniel, line 67:
```suggestion
            LicenseTexts = [.. licenseTexts],
```

PR #377, opened 2025-02-26
Proj0036: Remove None when redundant

### LINKED ISSUE #376: New Rule: <Content Include> does not require a <None Remove>
When changing the behavior of `<Content>`/`<None>` to `<Content>`, there is no need to have a remove `<None>` item. this reduces noise dramatically.

Corniel:
Closes #376

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Remove_None_when_redundant.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Remove_None_when_redundant.cs
@@ -1,0 +1,31 @@
+namespace Rules.MS_Build.Remove_None_when_redundant;
+
+public class Reports
+{
+    [Test]
+    public void on_not_alphabetical_order() => new RemoveNoneWhenRedundant()
+       .ForInlineCsproj(@"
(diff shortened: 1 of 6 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Remove_None_when_redundant.cs
```diff
@@ -0,0 +1,31 @@
...
+       .ForInlineCsproj(@"
+<Project Sdk=""Microsoft.NET.Sdk"">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include=""*"" />
+    <None Remove=""License.md"" />
```
CptWesley, line 16: Is this truely redundant? doesn't this include everything _except_ `License.md` or does it not work like that?

Corniel: It will have no influence on the output, as `<None>` has no influence. If hiding was intended, that can be solved otherwise, so yes, this should be redundant.

CptWesley (approved):

PR #379, opened 2025-02-26
Deal with custom third-party licenses

### LINKED ISSUE #337: License validation
After releasing [Pro:0500: Only include packages with an explicitly defined license](https://dotnet-project-file-analyzers.github.io/rules/Proj0500.html) (#335 ) It might be worth to think about a strategy on how to check for valid licenses. After some consideration I think it might be worth to split it up in multiple rules.

## Proj0501: Only include packages with a compliant license
This rule ignores licenses that are not understood (license files/urls that could not be converted to a known license). Licenses that violate the licenses specified in the `<AllowedLicenses>` are guarded, otherwise an **ERROR** is added.

## Proj0502: Only include packages with a reviewed license
This rules ignores rules that are compliant with Proj0501. If the `<AllowedThirdPartyPackages>` has no entry, an **ERROR** is added. To prevent unintended license changes, I though it might be a good idea to add an hash based on de agreed on license:

``` XML
<AllowedThirdPartyPackages>
   SonarAnalyzer.CSharp #L9ThxnotKPzthJ7hu3bnORuT6xI,
   MyCompany.*          #EiVUfR3CPFNgdDHFhpG5OcrzIYY
</AllowedThirdPartyPackages>
```

Obviously, this implies that the hash is communicated in the error messages.. Furthermore, a hash mismatch should lead to another error message, and may even a different diagnostic code.

Corniel:
For not-standard/custom third-party licenses, by adding `<ThirdPartyLicense Include="ID" Hash="{hash}" />` an custom license can be specified. 

**TODO**
- [x] Support wildcards in ID's
- [x] Support specifying versions when multiple versions of a package are supported with different licenses

Closes #337


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -783,8 +783,29 @@ public static partial class Rule
...
-        severity: DiagnosticSeverity.Error,
-        isEnabled: false);
+        severity: DiagnosticSeverity.Warning);
+
+    public static DiagnosticDescriptor CustomPackageLicenseUnknown => New(
+        id: 0503,
+        title: "Package license is unknown",
+        message: @"Add <ThirdPartyLicense Include=""{0}"" Hash=""{0}"" /> to accept the license",
+        description:
+            "To prevent legal issues do not rely on third-party references that have " +
```
CptWesley, line 793: description still needs to be changed I think

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -783,8 +783,29 @@ public static partial class Rule
...
+            "an incompatable license.",
+        tags: ["license"],
+        category: Category.Legal,
+        severity: DiagnosticSeverity.Warning);
+
+    public static DiagnosticDescriptor CustomPackageLicenseHasChanged => New(
+     id: 0504,
+     title: "Package license has changed",
+     message: "The license for {0} has changed as its hash now is {1}",
+     description:
```
CptWesley, line 803: description still needs to be changed I think

CptWesley (commented):
Perhaps it would also be a good idea to run the license text through the `PrepareLicenseText` method (or whatever the name was) to strip it of whitespace, special characters, etc. before hashing it, to make it less likely that a meaningless change to the license text (e.g. some whitespace or some capitalization) causes a different hash.

Corniel:
> Perhaps it would also be a good idea to run the license text through the `PrepareLicenseText` method (or whatever the name was) to strip it of whitespace, special characters, etc. before hashing it, to make it less likely that a meaningless change to the license text (e.g. some whitespace or some capitalization) causes a different hash.

That is already the case: 

``` C#
public static LicenseExpression FromFile(IOFile? licenseFile)
{
    if (licenseFile is not { Exists: true } file)
    {
        return Unknown;
    }

    var contentRaw = file.ReadAllText();
    var content = PrepareLicenseText(contentRaw);
    var contentNgrams = content.GetNGrams(LicenseTextLookupNGramSize);

    // Matching logic is loosely based on the logic used by Licensee: https://github.com/licensee/licensee
    // Licensee is the library used by GitHub for determining the license.

    foreach (var pair in LicenseTextLookup)
    {
        var modelNgrams = pair.Key;

        if (contentNgrams.DiceSorensenCoefficientAtLeast(modelNgrams, 0.95f))
        {
            return pair.Value;
        }
    }

    return CustomLicense.Create(content);
}
```

src/DotNetProjectFile.Analyzers/Rule.cs
Corniel: Yep. I'm already working on writing the documentation. The description will be adjusted based on that.

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -788,12 +788,33 @@ public static partial class Rule
         title: "Only include packages compliant with project",
         message: "The {0} package is distributed as {1}, which is imcompatable with the {2} license of the project",
         description:
-            "To prevent legal issues do not rely on third-party references that have " +
-            "an incompatable license.",
+            "To prevent legal issues do not rely on third-party references with " +
+            "custom defined license that has not been reviewed.",
```
CptWesley, line 792:
```suggestion
            "custom defined license that have not been manually reviewed.",
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -788,12 +788,33 @@ public static partial class Rule
...
-        isEnabled: false);
+        severity: DiagnosticSeverity.Warning);
+
+    public static DiagnosticDescriptor CustomPackageLicenseUnknown => New(
+        id: 0503,
+        title: "Package license is unknown",
+        message: @"Add <ThirdPartyLicense Include=""{0}"" Hash=""{0}"" /> to accept the license",
+        description:
+            "To prevent legal issues do not rely on third-party references with " +
+            "custom defined license that has not been reviewed.",
```
CptWesley, line 803:
```suggestion
            "custom defined license that have not been manually reviewed.",
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -788,12 +788,33 @@ public static partial class Rule
...
+        tags: ["license"],
+        category: Category.Legal,
+        severity: DiagnosticSeverity.Warning);
+
+    public static DiagnosticDescriptor CustomPackageLicenseHasChanged => New(
+     id: 0504,
+     title: "Package license has changed",
+     message: "The license for {0} has changed as its hash now is {1}",
```
CptWesley, line 811:
```suggestion
     message: "The license for {0} has changed as its hash is now {1}",
```

CptWesley:
I would also like to swap out the current hashing implementation (`SHA1Managed`) for the newer and faster `IncrementalHash` implementation of SHA256. I'll make a seperate PR for that tomorrow. Benchmarks:

| Method             | N | Mean       | Error    | StdDev   |
|------------------- |-- |-----------:|---------:|---------:|
| SHA1_Static        | 1 | 2,170.5 ns | 42.89 ns | 57.26 ns |
| SHA1_Object        | 1 | 2,026.7 ns | 20.98 ns | 16.38 ns |
| SHA1_Managed       | 1 | 2,038.8 ns | 36.23 ns | 33.89 ns |
| SHA1_Incremental   | 1 | 1,987.8 ns | 28.91 ns | 27.05 ns |
| SHA256_Static      | 1 | 1,022.6 ns | 13.42 ns | 11.21 ns |
| SHA256_Object      | 1 |   929.2 ns | 18.16 ns | 16.99 ns |
| SHA256_Managed     | 1 |   918.8 ns | 12.00 ns | 11.22 ns |
| SHA256_Incremental | 1 |   875.4 ns | 12.09 ns | 10.72 ns |

It might be unintuitive that SHA256 has better performance than SHA1, but modern hardware usually has optimization for computing SHA256 hashes (and Microsoft has more insentive to optimize) for the not-yet-cryptographically-broken SHA256.

Extra side-note: the current usage of the hashing is not thread-safe, so that should probably be fixed regardless.

Corniel:
> I would also like to swap out the current hashing implementation (`SHA1Managed`) for the newer and faster `IncrementalHash` implementation of SHA256.

Sure. I'd just wanted the hash not being to long.

CptWesley:
> Sure. I'd just wanted the hash not being to long.

SHA256 truncated to 128 bits is still considered secure and more collision resistant than SHA1, so I will do that to keep the hash small

CptWesley:
@Corniel The updated hashing PR that I'm about to open, so that the hashes in the tests don't change in case of any pre-release builds

CptWesley:
@Corniel See #382

Corniel:
@CptWesley merged #382 into this branch and updated the assertions.

CptWesley (approved):

PR #389, opened 2025-03-04
Proj0214: Severity as INFO as it is considered too opinionated

Corniel:


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -510,7 +510,8 @@ public static partial class Rule
             "disable package generation by defining the " +
             "<IsPackable> node with value 'false'.",
         tags: ["Configuration", "NuGet", "package"],
-        category: Category.Configuration);
+        category: Category.Configuration,
+        isEnabled: false);
```
CptWesley, line 514: Perhaps change the severity to `info` instead?

Corniel: I'm not sure how well that works. I can test that.

CptWesley: Doesn't disabling it mean it can never be enabled again? Setting it to `info` (or perhaps better `none`) makes it so that it doesn't trigger warnings, but can be enabled to still trigger warnings if people so desire

Corniel: It seems to work. Lets do that instead.

CptWesley (approved):

PR #395, opened 2025-03-04
Add dotnet as extra search tag

CptWesley:
I am bothered by the fact that searching for `dotnet project file analyzer` does not make the library appear on nuget, whereas it does appear when searching for `project file analyzer`. I hope adding `dotnet` as a tag will solve this.


Corniel (approved):
LGTM

PR #397, opened 2025-03-04
Change suggested analyzer logic to use packagereferences rather than assembly names

### LINKED ISSUE #391: Analyzer suggestion is based on assembly names rather than package names
Currently the analyzer suggestions are based on assembly names, rather than package names. This causes for example `AwesomeAssertions` to require `FluentAssertions.Analyzers` instead of `AwesomeAssertions.Analyzers` due to the fact that their assembly name is still `FluentAssertions`

CptWesley:
Closes #391

Also includes logic to resolve transitive dependencies from package references, which can be used for extending the license analyzers later on.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -73,7 +73,7 @@ ToBeReleased:
 - Proj0506: Third-party license registry requires hash. (NEW RULE)
 - Proj0507: Third-party license registry must be unconditional. (NEW RULE)
 - Proj1001: Suggested analyzers now based on package references rather than assembly names. (FP)
-- Proj1001: Now suggest to use AwesomeAssertions.Analyzers for AwesomeAssertions.
+- Proj1001: Now suggest to use AwesomeAssertions.Analyzers for AwesomeAssertions. (FN)
 - Proj1200: Fix compilation packages (that only hold a collective set of dependencies) incorrectly marked as development-dependencies. (FP)
 - Proj1701: Do not report on short (less then 3 lines) release notes. (FP)
 v1.5.10.1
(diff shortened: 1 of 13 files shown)


src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -72,6 +72,8 @@ ToBeReleased:
 - Proj0505: Third-party license registry requires include. (NEW RULE)
 - Proj0506: Third-party license registry requires hash. (NEW RULE)
 - Proj0507: Third-party license registry must be unconditional. (NEW RULE)
+- Proj1001: Suggested analyzers now based on package references rather than assembly names. (FP)
+- Proj1001: Now suggest to use AwesomeAssertions.Analyzers for AwesomeAssertions.
```
Corniel, line 76:
```suggestion
- Proj1001: Now suggest to use AwesomeAssertions.Analyzers for AwesomeAssertions. (FN)
```

src/DotNetProjectFile.Analyzers/NuGet/PackageCache.cs
```diff
@@ -124,12 +124,13 @@ private static string GetRawPath()
             Directory = versionDir,
             HasAnalyzerDll = HasDllFiles("analyzers"),
             HasRuntimeDll = HasDllFiles("lib") || HasDllFiles("runtimes"),
-            HasDependencies = nuspec?.Metadata.Depedencies.Length > 0,
-            IsDevelopmentDependency = nuspec?.Metadata.DevelopmentDependency,
+            HasDependencies = nuspec?.Metadata?.Depedencies?.SelectMany(d => d.Dependencies ?? [])?.Any() ?? false,
```
Corniel, line 127:
```suggestion
            HasDependencies = nuspec?.Metadata?.Depedencies?.SelectMany(d => d.Dependencies ?? [])?.Any() is true,
```

Corniel (approved):
LGTM!

PR #398, opened 2025-03-04
Add snapshot test for default severities

### LINKED ISSUE #361: Provide default `.globalconfig` in documentation
From a usability standpoint it would be nice if we can provide an example `.globalconfig` file with all rules and their default severity (and comment with rule name/message).

CptWesley:
Closes #361


Corniel (approved):
Besides my comment: LGTM!

PR #400, opened 2025-03-04
Simplify naming and types

Corniel:
`CachedPackage` as `Package`. The usage of the previous `Package` has been reduced to one rule defining which analyzers should be used given other dependencies.


src/DotNetProjectFile.Analyzers/MsBuild/PackageReferenceBase.cs
```diff
@@ -54,15 +54,15 @@ public abstract class PackageReferenceBase(XElement element, Node parent, MsBuil
     public string? ResolveVersion()
         => ResolveVersionVerbose()?.Version;
 
-    public CachedPackage? ResolveCachedPackage()
+    public Package? ResolveCachedPackage()
```
CptWesley, line 57: these method names can then also be simplified

CptWesley (approved):

PR #401, opened 2025-03-04
Check transitive depedencies too

### LINKED ISSUE #359: Licensing: Check transitive dependencies
Currently only first-class dependencies are validated, but for a better picture it would be wise to also check transitive dependencies (preferably in a smart way if transitive dependencies occur multiple times).

Corniel:
Checks dependencies and there transitive dependencies too.

Closes #359


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/ThirdPartyLicenseResolver.cs
```diff
@@ -24,50 +24,73 @@ protected override void Register(ProjectFileAnalysisContext context)
...
+        var queue = new Queue<Dependency>();
+        var done = new HashSet<PackageVersionInfo>();
+
         foreach (var reference in context.File.ItemGroups.SelectMany(g => g.Children)
             .OfType<PackageReferenceBase>()
-            .Where(r => r.Version is { Length: > 0 }))
+            .Where(r => r is { Info.Version.Length: > 0 } && done.Add(r.Info)))
         {
-            Report(reference, projectLicense, licenses, context);
+            queue.Enqueue(new(reference, reference.Info));
```
CptWesley, line 34: shouldn't this also be guarded by a `if(done.Add(...))` ? Otherwise you might still see some duplicates

Corniel: The `.Where` clause contains that constraint, doesn't it?

CptWesley (approved):

Corniel:
I'm not sure why the build fails on transitive dependencies that succeed locally. Investigation is needed.

CptWesley:
> I'm not sure why the build fails on transitive dependencies that succeed locally. Investigation is needed.

Could it be that it is resolving different package versions locally?

Corniel:
> Could it be that it is resolving different package versions locally?

What I've found so far: 
* Some packages can not be resolved on the server
* Some packages are resolved with different versions
* Some packages seem to have different license URL's

I'm considering to ignore the test, and create a per-release to see how this works for real-world projects.

PR #408, opened 2025-03-08
Proj0038: Specify NoWarn rule IDs fully

### LINKED ISSUE #406: New Rule: NoWarn rules should have prefix
When using `<NoWarn>`, it is allowed to define a rule by only the integer part of the rule. This makes it even harder to find out which rule is actually ignored.

Corniel:
Closes #406


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Specify_NoWarn_rule_IDs_fully.cs
```diff
@@ -0,0 +1,29 @@
...
+public class Reports
+{
+    [Test]
+    public void IDs_specified_as_ints() => new SpecifyNoWarnRuleIdsFully()
+        .ForInlineCsproj(@"
+<Project Sdk=""Microsoft.NET.Sdk"">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <NoWarn>$(NoWarn);1700;NuGet1700;42</NoWarn>
```
CptWesley, line 12: Perhaps change the numeric part of `NuGet1700` to be different than the initial `1700` to verify it doesn't trigger a warning due to the number being duplicated.

src/DotNetProjectFile.Analyzers/MsBuild/NoWarn.cs
```diff
@@ -0,0 +1,9 @@
+namespace DotNetProjectFile.MsBuild;
+
+public sealed class NoWarn(XElement element, Node parent, MsBuildProject project)
+    : Node<string>(element, parent, project) 
+{
+    public IReadOnlyList<string> RuleIds
```
CptWesley, line 6: Should we also trim after splitting? I'm not sure how whitespace is treated here by MSBuild

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -342,6 +342,16 @@ public static partial class Rule
         tags: ["bug"],
         category: Category.Bug);
 
+    public static DiagnosticDescriptor SpecifyNoWarnRuleIdsFully => New(
+        id: 0038,
+        title: "Specify NoWarn rule IDs fully",
+        message: "Specify rule ID '{0}' fully by adding its prefix",
+        description:
+            "Disabling a rule by only its numeric part is harder to understand " +
+            "and can lead to excluding mulitiple rules at once.",
```
CptWesley, line 351:
```suggestion
            "and can lead to unintentionally excluding mulitiple rules at once.",
```

src/DotNetProjectFile.Analyzers/MsBuild/NoWarn.cs
Corniel: It is more that we better detect the numbers correctly. So, yes, trimming is a good idea.

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -342,6 +342,16 @@ public static partial class Rule
         tags: ["bug"],
         category: Category.Bug);
 
+    public static DiagnosticDescriptor SpecifyNoWarnRuleIdsFully => New(
+        id: 0038,
+        title: "Specify NoWarn rule IDs fully",
+        message: "Specify rule ID '{0}' fully by adding its prefix",
```
CptWesley, line 348:
```suggestion
        message: "Full specify rule ID '{0}' by adding its prefix",
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -342,6 +342,16 @@ public static partial class Rule
         tags: ["bug"],
         category: Category.Bug);
 
+    public static DiagnosticDescriptor SpecifyNoWarnRuleIdsFully => New(
+        id: 0038,
+        title: "Specify NoWarn rule IDs fully",
```
CptWesley, line 347:
```suggestion
        title: "Fully specify NoWarn rule IDs",
```

CptWesley (approved):

PR #409, opened 2025-03-08
Use .Children<> instead of SelectMany

Corniel:
To reduce a lot of plumbing I've created a method on `Nodes<T>` to simplify the code:

``` C#
context.Files.ItemGroups
    .SelectMany(i => i.Children)
    .OfType<Usings>()
    .Where(condtion):
```

can now be written as:

``` C#
context.Files.ItemGroups.Children<Usings>(condtion):
```


CptWesley:
Perhaps you can go one step further and add a `Descendants<>` which does `Children<>` recursively so that you can do `context.Files.Descendants<Usings>(condtion)` or maybe even `context.Descendants<Usings>(condtion)`

Corniel:
> Perhaps you can go one step further and add a `Descendants<>` which does `Children<>` recursively so that you can do `context.Files.Descendants<Usings>(condition)` or maybe even `context.Descendants<Usings>(condition)`

In all cases (except one) it only involved the direct children, and that is also the common case. The type of the file is not perse XML so context.Deccendants is less trival (with extension methods it can be done, but for now, I think I like this improvement already a lot).

Issue #414, opened 2025-03-19
Support SLNX format


Corniel:
Consider both rules on the SLNX file (via `.net.csproj`) as potentially using the referenced projects in the SLNX to do further analysis.

https://devblogs.microsoft.com/visualstudio/new-simpler-solution-file-format/

PR #415, opened 2025-03-19
Adopt SLNX project file format

Corniel:
Now it's out there, it is nice to adopt this clean and shiny format.

See #414


CptWesley (approved):
LGTM. Do we also want to add it to the general XML style analyzers?

PR #417, opened 2025-03-23
Proj0039: Treat all warnings as errors is considered a bad practice

### LINKED ISSUE #405: New Rule: Warnings as errors is not good
Laterally, I've seen multiple code bases, and heard people advocating to treat all warnings as errors. The reasoning is that it should force people to write better code, which in itself is something to endorse.

However there some objections:

1. Some tooling (such as Visual Studio spelling checker, or Sonar Lint) who report warnings as well, will make that solutions will not longer build for developers how use this tooling
2. New versions of analyzers containing new rules might be hard to adopt once they contain rules that trigger warnings in an exiting code base. It forces to change exiting code. The alternative of ignoring such new rules makes that they can not be adopted for (new) code
3. Static code analysis rules might report false positives. In some cases that means the implementation of the rule has to change (bugfix). This takes time, and if we suppress the issue(s) it is not unthinkable that the suppression is never released, cluttering the code base
4. The easiest way to get rid of warnings is by suppressing them. So with this policy enabled, you will see a lot of suppressions, and workarounds that might solve the warning, but not improve the code quality.

Corniel:
Closes #405 

For documentation see: https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers.github.io/pull/109

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Treat_warnings_as_warnings.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Treat_warnings_as_warnings.cs
@@ -3,7 +3,7 @@ namespace Rules.MS_Build.Treat_warnings_as_warnings;
 public class Reports
 {
     [Test]
-    public void when_property_is_defined() => new TreatWarningsAsWarnings()
+    public void when_treat_all_warnings_is_enabled() => new TreatWarningsAsWarnings()
        .ForInlineCsproj(@"<Project Sdk=""Microsoft.NET.Sdk"">
 
   <PropertyGroup>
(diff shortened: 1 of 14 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Treat_warnings_as_warnings.cs
```diff
@@ -0,0 +1,26 @@
+namespace Rules.MS_Build.Treat_warnings_as_warnings;
+
+public class Reports
+{
+    [Test]
+    public void when_property_is_defined() => new TreatWarningsAsWarnings()
```
Corniel, line 6:
```suggestion
    public void when_treat_all_warnings_is_enabled() => new TreatWarningsAsWarnings()
```

PR #420, opened 2025-03-25
Release v1.5.12

Corniel:
- Proj0005: Fixed missing package name when reference was using Update instead of Include. (BUG)
- Proj0039: Treat all warnings as errors is considered a bad practice. (NEW RULE)
- Proj0244: Exclude development dependencies. (FP)
- Proj0809: Fixed missing package name when reference was using Update instead of Include. (BUG)
- Proj0810: Fixed missing package name when reference was using Update instead of Include. (BUG)
- Proj1200: Fixed missing package name when reference was using Update instead of Include. (BUG)


CptWesley (approved):

Corniel:
@renzo-baasdam This one seems to failing consistently. The generation of the SBOM does not seem the cause.

renzo-baasdam:
> @renzo-baasdam This one seems to failing consistently. The generation of the SBOM does not seem the cause.

For me still only fails the first build, then any subsequent builds succeeds. Reverting https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/pull/415 seems to work for me, so perhaps reverting that for now and try to see if reintroducing the slnx format gives any issues.

Corniel:
@renzo-baasdam As you can see, I now re-added the sln, but that did not fix the issue. :S

renzo-baasdam:
Disabling SBOM generation does now seem to work again (aside from a failing test).

CptWesley:
I ran into this issue before with some other pull request, but forgot what the actual issue was in the end...

PR #432, opened 2025-05-19
Investigate failing builds

Corniel:
It seems that while building, somethings goes wrong related to generating the SBOM or some other side process. This PR is find out what is causing this issue, and how to fix it.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -11,7 +11,7 @@ jobs:
       - name: Setup .NET 9.0
         uses: actions/setup-dotnet@v1
         with:
-          dotnet-version: "9.0.x"
+          dotnet-version: 9.0.203
       - name: Build with dotnet
         run: dotnet build dotnet-project-file-analyzers.slnx --configuration Release
       - name: Run unit tests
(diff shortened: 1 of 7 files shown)


renzo-baasdam:
Looks like the same issue as in https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/pull/420. Although the pipeline in that PR succeeded, the merge into main pipeline failed [again](https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/actions/runs/14109533698/job/39525336857). No clue what exactly causes this.

Corniel:
@renzo-baasdam Generating the package and then trying to move it seems to be the real problem. I also have the problem when the SBOM is not generated.

renzo-baasdam:
Do you mean the `- name: Copy .nupkg files to nuget_packges folder` step in the pipeline? Does removing this solve the issue?

renzo-baasdam:
I would indeed those steps not to be the issue here, as the `dotnet build` step itself tends to generate the error.

Corniel:
@renzo-baasdam This 'fixes' the build, be now publishing will be manual thing again. Further investigation is needed.

PR #435, opened 2025-05-19
Proj0401: Only publish executables

### LINKED ISSUE #403: New Rule: IsPublisable for other then exe
We have a rule that specifies that `IsPublisable` should be specified explicitly which is (I believe) still a great rule. However, although you can use publish on library projects, I still have found no argument why that would make sense in the first place. So we might want to suggest to turn it of for those project types.

Corniel:
Linked documentation: https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers.github.io/pull/111

Closes #403

### DIFF
--- a/src/DotNetProjectFile.Analyzers/MsBuild/Project.Property.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/Project.Property.cs
@@ -38,6 +38,14 @@ public sealed partial class Project
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     private bool? managePackageVersionsCentrally;
 
+    public OutputType.Kind GetOutputType()
+        => Property<OutputType>()?.Value
+        ?? Sdk switch
+        {
+            "Microsoft.NET.Sdk.Web" => OutputType.Kind.Exe,
+            _ => OutputType.Kind.Library,
+        };
+
     public TNode? Property<TNode>() where TNode : Node => this
         .SelfAndDirectoryProps()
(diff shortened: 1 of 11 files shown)


src/DotNetProjectFile.Analyzers/MsBuild/Project.Property.cs
```diff
@@ -38,6 +38,14 @@ public bool IsDevelopmentDependency()
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     private bool? managePackageVersionsCentrally;
 
+    public OutputType.Kind GetOutputType()
+        => Property<OutputType>()?.Value
+        ?? Sdk switch
+        {
+            "Microsoft.NET.Sdk.Web" => OutputType.Kind.Exe,
```
CptWesley, line 45: Can you add a test case for this SDK case?

Corniel: It a bit more work (as I can not rely on an inlineproject due to the fact that I have to define an main entrypoint, but I'll do it)

CptWesley: perhaps it's easier to alter the inlineproject setup to also always add an empty entrypoint, so that we can use it for other test cases in the future as well

Corniel: Might be an idea. I added a project for now, and that was also not too much of an hassle after all.

PR #436, opened 2025-05-20
Proj0042, Proj0043, Proj0044 Lock files

### LINKED ISSUE #25: New Rule: Use lock files
Use lock files to prevent unitended package changes.

See:
* https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/
* https://www.meziantou.net/faster-and-safer-nuget-restore-using-source-mapping-and-lock-files.htm
* 
https://www.gresearch.com/blog/article/restoring-nuget-packages-with-a-lock-file/
https://blog.inedo.com/nuget/manage-dependencies-lockfiles-package-consumers/
https://www.damirscorner.com/blog/posts/20220708-UsingNuGetWithPackagesLockJson.html
https://www.endorlabs.com/blog/strengthening-security-in-net-development-with-packages-lock-json

CptWesley:
Closes #25 

NB: the way of checking the conditions is not fool proof, but should cover the general case (best effort). Things would be easier if we ever have proper property evaluation.


README.md
```diff
@@ -3,7 +3,7 @@
 
 ![.NET project file analyzers logo](design/logo_128x128.png)
 # .NET project file analyzers
-Contains 120+ [Roslyn](https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/)
+Contains 130+ [Roslyn](https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/)
```
Corniel, line 6: Yeah!!!

projects/CompliantCSharpPackage/packages.lock.json
```diff
@@ -0,0 +1,25 @@
+{
```
Corniel, line 1: Does this file follow the .editorconfig preferences? (If not, we might want to exclude it from the BOM rule)

CptWesley: At the very least my editor is not complaining about it when I do a clean rebuild

specs/DotNetProjectFile.Analyzers.Specs/TestTools/CachedProjectLoader.cs
```diff
@@ -11,6 +11,7 @@ internal static class CachedProjectLoader
 
     private sealed class Entry(FileInfo file)
     {
+        private readonly Random rnd = new(file.GetHashCode());
```
Corniel, line 14: May be document what is going on here. I assume the magic is needed, but I do not know why.

Corniel (approved):
Besides some suggestions: LGTM!

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -79,6 +85,9 @@ v1.5.13
 - Proj0025: Change advise to the use of .globalconfig. (DOCUMENTATION)
 - Proj0040: Run NuGet security audits on transitive dependencies too. (NEW RULE)
 - Proj0041: NuGet security audits should report on moderate issues at minimum. (NEW RULE)
+- Proj0042: Enable <ContinuousIntegrationBuild> when running in CI pipeline. (NEW RULE)
```
Corniel, line 88: Move this block to release notes and bump the version: we're ready to ship

projects/CompliantCSharpPackage/packages.lock.json
CptWesley: They were created without BOM in any case

Corniel:
> Closes #25
> 
> NB: the way of checking the conditions is not fool proof, but should cover the general case (best effort). Things would be easier if we ever have proper property evaluation.

We can check the actual value of a property once we add `<CompilerVisibleProperty>` for a property. I was not able to set up unit tests yet that actually can read those values, but in theory it should work. I think that for now this is fine.

PR #437, opened 2025-05-22
Update build pipeline

pmdevers:

PR #438, opened 2025-05-22
Create and publish release on git tag

pmdevers:
Update the workflow to release on git tag starting with v (e.g. :  v1.0.0 or v1.0.0-alpha0001)

- runs `build` and `test`
- `pack` if it is a tag build

on tag builds
- Create a Github Release with release notes
- Marks the Github Release as prerelease if the tag contains a - like `v1.0.0-alpha0001` tags


.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,88 @@
...
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '9.0.300'
```
Corniel, line 30: what is the best way to keep track of newer versions?

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,88 @@
...
+    branches:
+      - main
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  NUGET_FEED: https://api.nuget.org/v3/index.json
+  NUGET_KEY: ${{ secrets.NUGET_API_KEY }}
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
```
Corniel, line 18: Should we fix this or not? I'm still not sure what is best.

Corniel:
@pmdevers Can you address @renzo-baasdam 's remarks as far as still relevant made in #437 ?

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,88 @@
+name: Build
+on: 
+  push:
+    branches:
+      - main
+    tags:
+    - 'v*'
```
CptWesley, line 7: is this indentation correct? I would have expected it to be on the same level as `branches` above

Gemberkoekje (approved):

.github/workflows/build-publish.yml
pmdevers: It can be changed to '9.0' so it always uses the latest patch version

.github/workflows/build-publish.yml
pmdevers: The indentation is correct but I will make it consisted

pmdevers:
> @pmdevers Can you address @renzo-baasdam 's remarks as far as still relevant made in #437 ?

The issues mentioned by @renzo-baasdam are not relevant anymore commands are single lined

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: DotNet Pack
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          arrTag=(${GITHUB_REF//\// })
+          VERSION="${arrTag[2]}"
+          echo Version: $VERSION
+          VERSION="${VERSION//v}"
+          echo Clean Version: $VERSION
+          dotnet pack dotnet-project-file-analyzers.slnx -v normal --no-build --no-restore -p:PackageVersion=$VERSION -c Release --output packages
```
renzo-baasdam, line 46: Does the version of the tag now determine the package version? So the version number is the csproj's are then not used anymore?

Corniel: I'm not sure if that is desired. @pmdevers Why do you think that is better?

renzo-baasdam:
Not sure why, but the publish job also has a check for this PR.

.github/workflows/build-publish.yml
pmdevers: It is considered best practice in GitOps and CI/CD Workflows.

1. Ensures Traceability
   - a Tag points to a specific commit, using it as a version ensures it can reproduce the exact same code version
   - it gives a direct immutable reference to the source code that produced the package
2. Simplifies Release Management
   - You can automate versioning in build pipelines without manual intervention
   - Reduce human error and aligns the package version with the repository release history
3. Avoids Version Dirft
   - If the version is hardcoded or manually managed, it van fall out of sync with the repository state.
   - Deriving it from Git ensures consistency between the version label and the actual code contents

.github/workflows/build-publish.yml
Corniel: @pmdevers I know. And `9.0.x` is also possible. The question is, if we thing about predictable builds, what is considered best these days?

.github/workflows/build-publish.yml
Corniel: But if you make a typo in your tag, you're screwed are you not?

pmdevers: You can remove a tag without code commit

Corniel: sure, but if the package build is running you might be too late, are you not?

CptWesley, line 46:
```suggestion
          dotnet pack dotnet-project-file-analyzers.slnx -v normal --no-build --no-restore -p:Version=$VERSION -c Release --output packages
```

In this case I think I would be in favor of the command overwriting the `Version` property, rather than the `PackageVersion` (the default of `PackageVersion` is using the value of `Version`).

pmdevers: That is the case either way but to change the version you will have to create a new commit

.github/workflows/build-publish.yml
pmdevers: This is the way.

https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-net

Only discussion is if you want to pin the patch or minor version

.github/workflows/build-publish.yml
pmdevers: I can agree with this

Corniel (approved):
LGTM!

PR #454, opened 2025-06-03
Support SLNX files

Corniel:
With this PR, it is now possible to write rules on SLNX files too.


src/DotNetProjectFile.Analyzers/Extensions/Microsoft.CodeAnalysis.Diagnostics.AnalysisContext.cs
```diff
@@ -78,4 +78,16 @@ public static void RegisterResourceFileAction(this AnalysisContext context, Acti
             }
         });
     }
+
+    /// <summary>Registers an action on <see cref="ProjectFileAnalysisContext"/>.</summary>
```
CptWesley, line 82:
```suggestion
    /// <summary>Registers an action on <see cref="SolutionFileAnalysisContext"/>.</summary>
```

PR #464, opened 2025-06-16
Fix Proj0240 message incorrectly suggesting to disable DevelopmentDependency, rather than to enable it

CptWesley:
Development dependencies most likely don't need to have a consistent API, so that would make more sense. The analyzer implementation already behaves like this, so it was most likely a typo or copy-paste error in the message text.


Corniel (approved):
LGTM!

PR #465, opened 2025-06-17
Proj1001 Fix false-positive suggestion to use Microsoft.CodeAnalysis.Analyzers

### LINKED ISSUE #421: FP Proj1001: Analyzers should be needed due to other analyzers being included
If we look to this MS-Build file:

``` XML
<Project>

  <ItemGroup Label="Analyzers">
    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" PrivateAssets="all" />
  </ItemGroup>

</Project>
````

It reports:
> Use Microsoft.CodeAnalysis.Analyzers to analyze Microsoft.CodeAnalysis.NetAnalyzers

That seems not needed.

### PR DESCRIPTION
**Pull request by CptWesley**
Closes #421

### DIFF
--- a/src/DotNetProjectFile.Analyzers/NuGet/Package.cs
+++ b/src/DotNetProjectFile.Analyzers/NuGet/Package.cs
@@ -29,4 +29,6 @@ public sealed record Package
     public required LicenseExpression License { get; init; }
 
     public required NuSpecFile? NuSpec { get; init; }
+
+    public bool IsAnalyzerOnly => HasRuntimeDll && !HasRuntimeDll;
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/DotNetProjectFile.Analyzers/NuGet/Package.cs [hunk 1]
```diff
@@ -29,4 +29,6 @@ public sealed record Package
     public required LicenseExpression License { get; init; }
 
     public required NuSpecFile? NuSpec { get; init; }
+
+    public bool IsAnalyzerOnly => HasRuntimeDll && !HasRuntimeDll;
```
Corniel, line 33: Is this corrrect?

CptWesley: No

Corniel (approved):
LGTM!

PR #472, opened 2025-06-19
Add parent package to message for issues with transitive licenses

LauraXiulan:
Add the parent package to the warning message when having issues with the license of transitive packages to make it more clear where the issue is. Also fixed some typos. Updated test to include a nested transitive package.


Corniel (approved):
Don't forget to update the release notes
LGTM!

PR #492, opened 2025-08-11
Proj0509: NuGet package cache could not be resolved

Corniel:
As we want to be informed if none of the license checking rules can be performed.


docs/rules/Proj0509.md
```diff
@@ -0,0 +1,9 @@
+---
+parent: General
+ancestor: MSBuild
+permalink: /rules/Proj0509
+---
+
+# Proj0509: NuGet package cache could not be resolved
+The rules that check for licenses need (read) access to the package cache. If
+that location could not be resolved, none of those rules works properly.
```
LauraXiulan, line 9:
```suggestion
that location could not be resolved, none of those rules work properly.
```

PR #493, opened 2025-08-12
Proj3000: Ignore packages.lock.json files

Corniel:
When generating lock files (`packages.lock.json`), Microsoft ignores preferences defined in the `.editorconfig`, hence on every new build those file get an UTF-8 BOM header (back). Reporting on this file is therefor considered a false positive and excluded as a result.

### DIFF
--- a/projects/MultipleEncodings/packages.lock.json
+++ b/projects/MultipleEncodings/packages.lock.json
@@ -1,0 +1,13 @@
+﻿{
+  "version": 2,
+  "dependencies": {
+    "net9.0": {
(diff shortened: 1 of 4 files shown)


Corniel:
> Actually, I just checked and in my setup it does seem like it respects the editorconfig when generating the lock files. Are you sure your editorconfig covers .json files?

Hmm, let me check again...

Corniel:
It turned out to be `CompatibilitySuppressions.xml` not `packages.lock.json`...

PR #497, opened 2025-08-17
Proj5005: Omit Project ID's

### LINKED ISSUE #425: New Rule: ID attribute in SLNX seems to have no purpose
In the brand new SLNX solution files, it seems that when converting it, the `@ID` attribute is set sometimes, but that does not seems to have any value.

Corniel:
Closes #425 

Creating tests for SLNX files should be as simple as creating tests for CS-Project files. Because that is not the case I left the unit test out for now, and created an issue to solve that issue (separately).

### DIFF
--- a/docs/rules/Proj5005.md
+++ b/docs/rules/Proj5005.md
@@ -1,0 +1,25 @@
+---
+parent: SLNX
+ancestor: Rules
+permalink: /rules/Proj5005
+---
+
+# Proj5005: Omit Project ID's
+Projects in the SLN file had an (unique) ID. In the SLNX file, this ID does not
+have any added value, and is considered noise.
+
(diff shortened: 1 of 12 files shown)


docs/rules/Proj5005.md
```diff
@@ -0,0 +1,25 @@
+---
+parent: SLNX
+ancestor: Rules
+permalink: /rules/Proj5005
+---
+
+# Proj5005: Omit Project ID's
+Projects in the SLN file had an (unique) ID. In the SLNX file, this ID does not
```
CptWesley, line 8:
```suggestion
Projects in the SLN file had a (unique) ID. In the SLNX file, this ID does not
```

PR #512, opened 2025-08-19
Proj1300: Project references should be compliant

### LINKED ISSUE #378: New Rule: Disallow certain project references
After a fruitful discussion with a colleague, about enforcing dependencies between projects, I thought we can potentially enforce a certain dependency structure based on project types.

Types that we can detect without extra's:
* TestProject (via `<IsTestProject>`)
  - Include: everything besides other test projects
* Web Applications (via `<IsPackable>`)
  - Include: Packages and Other
* Packages (via `<IsPackable>`)
  - Include: Packages
* Applications (via `<OutputType>`)
  - Include: Packages and Other
* Other
  - Include: Packages and Other

We could consider to support some kind of modules hat further restrict hits, but I think this set of limitations could be a good start.

Corniel:
The documentation/naming is not far from what I have envisioned, but the first set of rules has been implemented:

1. Test projects should not be included
2. Executables should not be included
3. Packables should only inlcude packables

Closes #378

### DIFF
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -83,7 +83,7 @@ v.1.7.2
 - Proj0219: Packable projects should be libraries. (NEW RULE)
 - Proj1300: Avoid project references to executables. (NEW RULE)
 - Proj1301: Project references of packable targets should be packable. (NEW RULE)
-- Proj1302: Avoid project references to test project. (NEW RULE)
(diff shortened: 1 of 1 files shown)


CptWesley:
Would it perhaps be worth it to add these checks as seperate rules? That way they can be disabled/surpressed seperately.

docs/rules/Proj1300.md
```diff
@@ -0,0 +1,8 @@
+---
+parent: Design
+ancestor: MSBuild
+permalink: /rules/Proj1300
+---
+
+# Proj1300: Project references should be compliant
+TODO
```
Corniel, line 8: TODO Has still be solved

PR #528, opened 2025-09-15
Proj0048: Language version should be set to an explicit version number

### LINKED ISSUE #522: Suggest <LangVersion>Preview</LangVersion> should not be used
As building application for customers i think we should not use 
`<LangVersion>Preview</LangVersion>`

But on `<LangVersion>latest</LangVersion>` or a specific version

CptWesley:
Closes #522

### DIFF
--- a/docs/rules/Proj0048.md
+++ b/docs/rules/Proj0048.md
@@ -1,0 +1,48 @@
+---
+parent: Packaging
+ancestor: Rules
+permalink: /rules/Proj0048
+---
+
+# Proj0048: Language version should be set to an explicit version number
+When `<LangVersion>` is not explicitly defined, the language version used by the compiler will depend on the language version associated with the target framework ([more info](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version)).
+This can result in unintentional side effects when targeting multiple target frameworks or when changing the targert framework version.
+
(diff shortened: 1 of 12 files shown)


docs/rules/Proj0048.md
```diff
@@ -0,0 +1,48 @@
...
+This can result in unintentional side effects when targeting multiple target frameworks or when changing the targert framework version.
+
+When `<LangVersion>` is set to `latest`, `latestMajor`, `default` or `preview`, [the actually used language version may be different depending on the .NET compiler version](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version). This means that for different members in the same team, the used language version may be different depending on which .NET versions they have installed locally. This results in builds that are harded to reproduce.
+
+To ensure the language version used by the compiler is not unintentionally changed, it is adviced to explicitly define the `<LangVersion>` node with an explicit version number. For a list of valid `LangVersion` values, please consult [this](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) list.
```
CptWesley, line 13:
```suggestion
To ensure the language version used by the compiler is not unintentionally changed, it is advised to explicitly define the `<LangVersion>` node with an explicit version number. For a list of valid `LangVersion` values, please consult [this](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) list.
```

PR #529, opened 2025-09-15
Inline any set of files as an analyzable project

### LINKED ISSUE #498: Create helpers to simplify writing tests on SLNX
Something like:

``` C#
    [Test]
    public void TODO_comment() => new SlnxAnalyzers()
      .ForInlineSlnx("{SLNX file}")
      .HasNoIssues();
```

CptWesley:
Closes #498

- Adds a builder for inline projects that allows attaching any files to it that will also be included in the analysis.
- Rewritten `AddInlineCsproj` to use new builder under the hood
- Added `AddInlineSdkProject` to quickly setup a builder with a valid sdk project.
- Added `AddInlineSlnx` to quickly setup an sdk project and include the given slnx content.

Included 1 redundant test case for slnx indentation to demonstrate the syntax.


Corniel (approved):
LGTM!

PR #540, opened 2025-09-22
Introduction of Language SVO

Corniel:
Introduction of Language SVO

### DIFF
--- a/src/DotNetProjectFile.Analyzers/CodeAnalysis/Language.cs
+++ b/src/DotNetProjectFile.Analyzers/CodeAnalysis/Language.cs
@@ -10,48 +10,54 @@ public readonly struct Language : IEquatable<Language>
     [DebuggerBrowsable(DebuggerBrowsableState.Never)]
     private readonly string Id;
 
-    private Language(string id) => Id = id;
+    private Language(string id, string name, string xxproj, bool roslynBased)
+    {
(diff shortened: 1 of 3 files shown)


src/DotNetProjectFile.Analyzers/CodeAnalysis/Language.cs
```diff
@@ -0,0 +1,88 @@
...
+    public static readonly Language CSharp = new(nameof(CSharp));
+
+    /// <summary>Visual Basic (.NET).</summary>
+    public static readonly Language VisualBasic = new(nameof(VisualBasic));
+
+    /// <summary>F#.</summary>
+    public static readonly Language FSharp = new(nameof(FSharp));
+
+    /// <summary>Gets the MSBuild project file extenion for the language.</summary>
+    public string? ProjectFile => Id switch
```
CptWesley, line 28: Would it maybe be better to have these fields/props be configured through the constructor instead?

CptWesley (approved):
2 Comments, otherwise looks fine

src/DotNetProjectFile.Analyzers/CodeAnalysis/Language.cs
Corniel: Done.

PR #557, opened 2025-10-03
Additional files are automatically added

Corniel:
Additional files are automatically added for a while now, adding files explicitly is not longer needed.


CptWesley:
Isn't this only the case when using the SDK?

PR #563, opened 2025-10-06
Automatically add the project file to the additional files

Corniel:
The pack instruction was missing...


props/common.props
```diff
@@ -20,10 +20,10 @@
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
   </PropertyGroup>
 
-  <ItemGroup>
+  <!--<ItemGroup>
```
CptWesley, line 23: Is this commenting intentional?

PR #591, opened 2025-11-14
Renaming to prevent issues when building code fixes

Corniel:
Renaming `Project` too.

See #590


Corniel:
@CptWesley Do you understand why these tests start failing? They still succeeded for #590

PR #599, opened 2025-12-10
Update to .NET 10

Corniel:
Build and tests.


CptWesley:
Buildalyzer seems to crash

PR #606, opened 2025-12-29
Proj0201: Version is not required when using NuGet.Versioning or MinVersion

### LINKED ISSUE #602: Proj0201 and MinVer
MinVer and some other packages allow automatic versioning without borther with manual versioning process changing `<Version>`  property in sources. I suggest to throw Proj0201 only if project does not use such packages (at least MinVer). For now Proj0201 is always false positive for such projects.

Corniel:
Closes #602


CptWesley:
@Corniel it should. inline projects get written to a temporary directory and then just analyzed like any other project, so there shouldn't be any behavioural differences.

Corniel:
> @Corniel it should. inline projects get written to a temporary directory and then just analyzed like any other project, so there shouldn't be any behavioral differences.

I know, but for these examples added, it does?!

Corniel:
> @Corniel does it work if you do it through a test project instead of the inline option?

Yep. I converted one in the last update.

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -79,6 +79,7 @@
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
 v1.8.1
+- Proj0201: <Version> is not required when using NuGet.Versioning or MinVersion. (FP)
```
sungam3r:
```suggestion
- Proj0201: <Version> is not required when using NuGet.Versioning or MinVer. (FP)
```

PR #619, opened 2026-01-03
Define <RepositoryUrl> is not needed for SDK's 8.0 and up

### LINKED ISSUE #601: Proj0205 and SourceLink
Hi. Source Link is enabled by default when building projects using the .NET 8 SDK or later. Setting `PublishRepositoryUrl` property set to `true` allows to automatically set `RepositoryUrl` property as well. I suggest to throw Proj0205 only for tfm < 8.0 and only when `PublishRepositoryUrl` property is not `true`.

Corniel:
Only checks [Proj0205](https://dotnet-project-file-analyzers.github.io/rules/Proj0205.html) when the SDK could not be detected, or predates .NET 8.0.

Closes #601

### DIFF
--- a/docs/rules/Proj0205.md
+++ b/docs/rules/Proj0205.md
@@ -14,7 +14,7 @@ defining the `<IsPackable>` node with value `false`.
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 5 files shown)


Corniel:
> Are we sure the SDK version is the actual thing that makes the difference? For other changes in default values they usually pick the highest targeted framework version.

Good question. I checked that without the URL set, it was in the NuGet package. I did not check if that would still be the case when only targeting .NET 6.0 (or another older TFM).

PR #620, opened 2026-01-04
Proj0049: Prefer the latest released language version

### LINKED ISSUE #527: Pefer newer language versions
The highest supported language is bounded to the available runtime, and we only can guess that based on the highest used `<TargetFramework(s)>` (potentially via the `.net.csproj`). Using an older version than the default version of a framework seems something that should be discouraged.

Somewhat related to #522

Corniel:
# Proj0049: Prefer the latest released language version 

To benefit from the latest language features, the latest released version of the language (if it is C#, or V.B.NET) should be used.

| .NET |   C# | VB.NET |
|-----:|-----:|-------:|
|  6.0 | 10.0 |  16.9  |
|  7.0 | 11.0 |  16.9  |
|  8.0 | 12.0 |  16.9  |
|  9.0 | 13.0 |  17.13 |
| 10.0 | 14.0 |  17.13 |

Closes #527

### DIFF
--- a/docs/rules/Proj0205.md
+++ b/docs/rules/Proj0205.md
@@ -14,7 +14,7 @@ defining the `<IsPackable>` node with value `false`.
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <IsPackable>true</IsPackable>
   </PropertyGroup>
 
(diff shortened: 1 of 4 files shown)


src/DotNetProjectFile.Analyzers/MsBuild/LanguageVersion.cs
```diff
@@ -0,0 +1,94 @@
...
+/// <summary>.NET SDK version.</summary>
+/// <remarks>
+/// See: https://learn.microsoft.com/en-us/dotnet/core/versions.
+/// </remarks>
+[TypeConverter(typeof(Converter))]
+public readonly record struct LanguageVersion(int Major, int Minor = 0) : IComparable<LanguageVersion>
+{
+    public static readonly LanguageVersion None;
+
+    public static readonly LanguageVersion Default = new(int.MaxValue, 0);
```
CptWesley: are the minor versions here magic values?

PR #621, opened 2026-01-04
Proj0808: Define CPM related configuration only in Directory.Packages.props

### LINKED ISSUE #548: CPM related config should only be defined in Directory.Packages.props
Items like `<PackageVersion>` and all related properties (except `<ManagePackageVersionsCentrally>` when the value is set to `false`) should only occur in the `Directory.Packages.props` file.

Corniel:
Actually, this is not creating a new rule, but extending an existing one.

Closes #548


Corniel:
@CptWesley I tested it: it works also outside `Directory.Packages.props`. It is not imported if you have a reference to a project file (which makes sense), but other shared props files act the same. Therefor I think it would be best to have one rule for all configuration to be in `Directory.Packages.props` and not having a separate rule for `<PackageVersion>` .

CptWesley (approved):

PR #633, opened 2026-01-10
Proj0049: Prefer the latest released language version for F#

Corniel:
F# support was lacking for this rule.


Corniel:
@sungam3r With your suggestion applied, okay now?

PR #634, opened 2026-01-11
Proj0219: Packable projects could be .NET CLI tools too

### LINKED ISSUE #631: Proj0219 false positive
```
The <IsPackable> node is set to 'true' but the <OutputType> is not 'Library' (https://dotnet-project-file-analyzers.github.io/rules/Proj0219.html)
```

This one is FP for .NET tools (.NET global/local tools). See https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create.

OutputType for such tools is `Exe`.

Examples:
- https://github.com/danielpalme/ReportGenerator/blob/main/src/ReportGenerator.DotnetGlobalTool/ReportGenerator.DotnetGlobalTool.csproj#L4
- https://github.com/PublicApiGenerator/PublicApiGenerator/blob/master/src/PublicApiGenerator.Tool/PublicApiGenerator.Tool.csproj#L4

Corniel:
Closes #631

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/PreferLatestLanguageVersion.cs
+++ b/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/PreferLatestLanguageVersion.cs
@@ -7,7 +7,7 @@ public sealed class PreferLatestLanguageVersion() : MsBuildProjectFileAnalyzer(R
     /// <inheritdoc />
     public override ImmutableArray<ProjectFileType> ApplicableTo => ProjectFileTypes.ProjectFile;
 
-    public override ImmutableArray<Language> ApplicableLanguages { get; } = [Language.CSharp, Language.VisualBasic];
+    public override ImmutableArray<Language> ApplicableLanguages { get; } = [Language.CSharp, Language.VisualBasic, Language.FSharp];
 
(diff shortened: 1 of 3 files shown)


docs/rules/Proj0219.md
```diff
@@ -6,7 +6,12 @@ permalink: /rules/Proj0219
 
 # Proj0219: Packable projects should be libraries
 Creating [NuGet packages](../general/nuget-packages.md) for project files other
-than libraries is so rare that we consder it to be a mistake.
+than libraries is unusual and often indicates a misconfiguration.
+
+A known exception to this rule are [.NET CLI tools](https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create)
+packages which requires `<PackAsTool>` to be `true`.
+
+This rule reports on any project that is packable and not a library or a .NET CLI tool.
```
LauraXiulan:
```suggestion
A known exception to this rule are [.NET CLI tools](https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create) packages. Because `<PackAsTool>` needs to be on `true` for this type of project, it can be excluded from the rule.

This rule reports on any project that is packable and not a library or an executable with the `<PackAsTool>` property on `true`.
```

PR #713, opened 2026-04-10
SonarQube integration should not add <Content> twice

### LINKED ISSUE #712: NETSDK1022:  Duplicate 'Content' items were included. When a web.config or app.config is included in the project directory
Version: 1.10

When adding version 1.10 to a AspNet Core Project, a service, I receive the error NETSDK1022 when compiling the project.

NETSDK1022: Duplicate 'Content' items were included. The .NET SDK includes 'Content' items from your project directory by default. You can either remove these items from your project file, or set the 'EnableDefaultContentItems' property to 'false' if you want to explicitly include them in your project file. For more information, see https://aka.ms/sdkimplicititems. The duplicate items were: 'web.config'

This seems to happen whenever there is a web.config or app.config included in the project directory.

Note:
With version 1.9.0 all compiles just fine.

Corniel:
Should fix #712

### DIFF
--- a/src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
+++ b/src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
@@ -2,15 +2,23 @@
 
   <!--
     SonarQube can pick up issues report on files others than compiling files
-    if they are reported as <Content>. To reduce noise and overhead, copying
-    to output is disabled and visibility is false.
+    if they are reported as <Content>.
+    
+    * Exclude="@(Content)" 
+      Prevents <Content> from being added twice.
+      
+    * CopyToOutputDirectory="never"
+      Ensures that <Content> only added to satisfy SonarQube is not copied.
+    
+    * Visible="false"
+      Prevents from <Content> being shown it he IDE that was included by the user.
   -->
   <ItemGroup Condition="'$(SonarQubeIntegration)' == 'true'">
-    <Content Include="**/*.??proj" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.config" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.props" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.resx" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.slnx" CopyToOutputDirectory="never" Visible="false" />
+    <Content Exclude="@(Content)" CopyToOutputDirectory="never" Visible="false" Include="**/*.??proj" />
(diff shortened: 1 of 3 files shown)


src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
```diff
@@ -2,15 +2,23 @@
...
     SonarQube can pick up issues report on files others than compiling files
-    if they are reported as <Content>. To reduce noise and overhead, copying
-    to output is disabled and visibility is false.
+    if they are reported as <Content>.
+    
+    * Exclude="@(Content)" 
+      Prevents <Content> from being added twice.
+      
+    * CopyToOutputDirectory="never"
+      Ensures that <Content> only added to satisfy SonarQube is not copied.
```
Corniel:
```suggestion
      Ensures that <Content> only included to satisfy SonarQube is not copied.
```

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
```diff
@@ -2,15 +2,23 @@
...
+    * Visible="false"
+      Prevents from <Content> being shown it he IDE that was included by the user.
   -->
   <ItemGroup Condition="'$(SonarQubeIntegration)' == 'true'">
-    <Content Include="**/*.??proj" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.config" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.props" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.resx" CopyToOutputDirectory="never" Visible="false" />
-    <Content Include="**/*.slnx" CopyToOutputDirectory="never" Visible="false" />
+    <Content Exclude="@(Content)" CopyToOutputDirectory="never" Visible="false" Include="**/*.??proj" />
```
CptWesley, line 17: This doesn't exclude all previously included content?

CptWesley (approved):

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
Corniel: Yes is does: So this Include statement excluded everything that is already content, which prevents things from being updated twice.

CptWesley: So the exclude is an exclude on the include?

Corniel: Yep.

PR #718, opened 2026-04-12
Build bug

Corniel:
Hopefully solved and closes #706

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/TestTools/InlineProjectAnalyzerVerifyContextBuilder.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/TestTools/InlineProjectAnalyzerVerifyContextBuilder.cs
@@ -9,6 +9,7 @@ namespace Specs.TestTools;
 
 internal sealed class InlineProjectAnalyzerVerifyContextBuilder
 {
+    private static readonly Lock Locker = new();
     private readonly DiagnosticAnalyzer Analyzer;
     private readonly ImmutableArray<FileDefinition> Files;
     private readonly Lazy<string> Hash;
@@ -110,7 +111,6 @@ internal sealed class InlineProjectAnalyzerVerifyContextBuilder
         }
     }
 
-    private static readonly Lock Locker = new Lock();
 
     private sealed record FileDefinition
     {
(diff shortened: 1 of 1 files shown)


specs/DotNetProjectFile.Analyzers.Specs/TestTools/InlineProjectAnalyzerVerifyContextBuilder.cs
```diff
@@ -75,37 +76,42 @@ private static FileDefinition ToFile(string name, string? content)
...
         {
             throw new InvalidOperationException("Requires at least 1 file.");
         }
 
-        var tempDir = Path.Combine(Path.GetTempPath(), "dotnet-project-file-analyzer/tests");
-        var dir = Path.Combine(tempDir, hash.Value);
-        Directory.CreateDirectory(dir);
-
-        foreach (var file in files)
+        lock (Locker)
```
CptWesley, line 89: This can still fail on file locks if the compiler has a lock on a file as well. Perhaps we should put the reads and writes in a try-catch loop until it succeeds?

specs/DotNetProjectFile.Analyzers.Specs/TestTools/InlineProjectAnalyzerVerifyContextBuilder.cs
```diff
@@ -75,37 +76,42 @@ private static FileDefinition ToFile(string name, string? content)
...
-        var fileInfo = new FileInfo(firstFileName);
+            var firstFileName = Path.Combine(dir, Files[0].Name);
+            var fileInfo = new FileInfo(firstFileName);
 
-        return ProjectFileAnalyzersDiagnosticAnalyzerExtensions.ForTestProject(analyzer, fileInfo);
+            return ProjectFileAnalyzersDiagnosticAnalyzerExtensions.ForTestProject(Analyzer, fileInfo);
+        }
     }
 
+    private static readonly Lock Locker = new Lock();
```
CptWesley: I would move this to the top of the file somewhere, rather than the bottom

specs/DotNetProjectFile.Analyzers.Specs/TestTools/InlineProjectAnalyzerVerifyContextBuilder.cs
Corniel: I would try to first see if issues reduce without having a retry mechanism in place.

PR #742, opened 2026-04-19
Proj0813: Global package references only work when CPM is enabled

Corniel:
While working on something else, I stumbled on the following:
``` xml
<ItemGroup Condition="'$(ManagePackageVersionsCentrally)' == 'true' And '$(RestoreEnableGlobalPackageReference)' != 'false'" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <!--
        Add GlobalPackageReference items to the PackageReference item group with no version.

        Global package references only include the same assets as a development dependency (runtime; build; native; contentfiles; analyzers)
        because those kind of packages are the best candidate for a global package reference.  They are generally packages that
        extend the build.

        Global package references have all assets private because central package references are generally packages that provide
        versioning, signing, etc and should not flow to downstream dependencies.  Also, central package references are already
        referenced by every project in the tree so they don't need to be transitive.
      -->
    <PackageReference Include="@(GlobalPackageReference)" Version="" IncludeAssets="Runtime;Build;Native;contentFiles;Analyzers" PrivateAssets="All" />
    <!--
        Add GlobalPackageReference items to the PackageVersion item group with the version.
      -->
    <PackageVersion Include="@(GlobalPackageReference)" Version="%(Version)" />
  </ItemGroup>
```

This is how MSBuild implemented `<GlobalPackageReference>`'s. It is obvious from this, that the latter is ignored when CPM is not enabled. Hence this rule. As part of the implementation, I primary rely on the value of `$(ManagePackageVersionsCentrally)`, and only when that is not provided, fall back to walking to the XML. This effected all other CPM rules.


Frulfump:
Makes sense, the docs are a bit unclear https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

But say
> GlobalPackageReference items should be placed in your Directory.Packages.props to be used by every project in a repository:

And that's traditionally a CPM file but It doesn't say MUST be placed, and I found no other official docs discussing GlobalPackageReferences, but it would have been nice if the docs contained a note stating that it's only for CPM.

The blog post where CPM was introduced https://devblogs.microsoft.com/dotnet/announcing-nuget-6-4-signed-central-delivered/

Says
> A new feature to central package management is <GlobalPackageReference>...

So that more clearly relates it to CPM.

Seems like GlobalPackageReference was a thing before CPM but only for a specific "custom" MSBuild SDK https://github.com/microsoft/MSBuildSdks/tree/main/src/CentralPackageVersions#global-package-references

Refs:
https://github.com/NuGet/Home/issues/10159
https://github.com/NuGet/Home/pull/11794
https://github.com/NuGet/NuGet.Client/pull/4680

Sidenote: I get a 404 for the XMLNS http://schemas.microsoft.com/developer/msbuild/2003 after it redirects to https://schemas.microsoft.com/developer/msbuild/2003
> The resource you are looking for has been removed, had its name changed, or is temporarily unavailable.

CptWesley (approved):

specs/DotNetProjectFile.Analyzers.Specs/Rules/Global_package_references_only_work_with_CPM.cs
```diff
@@ -0,0 +1,48 @@
...
+          </PropertyGroup>
+        
+          <ItemGroup>
+            <GlobalPackageReference Include="NuGet.Versioning" Version="7.3.1" />
+          </ItemGroup>
+        
+        </Project>
+        """)
+        .HasIssue(
+            Issue.WRN("Proj0813", "Global package reference 'NuGet.Versioning' is ignored as CMP is disabled").WithSpan(7, 4, 7, 73));
```
Frulfump: Nit: Typo
```suggestion
            Issue.WRN("Proj0813", "Global package reference 'NuGet.Versioning' is ignored as CPM is disabled").WithSpan(7, 4, 7, 73));
```

src/DotNetProjectFile.Analyzers/Extensions/DotNetProjectFile.Diagnostics.ProjectFileAnalysisContext.cs
```diff
@@ -0,0 +1,13 @@
+namespace DotNetProjectFile.Diagnostics;
+
+internal static class DotNetProjectFile
+{
+    extension(ProjectFileAnalysisContext context)
+    {
+        /// <summary>Indicates that CMP is enabled.</summary>
```
Frulfump: Nit: Typo
```suggestion
        /// <summary>Indicates that CPM is enabled.</summary>
```

src/DotNetProjectFile.Analyzers/MsBuild/MsBuildProps.cs
```diff
@@ -15,6 +15,9 @@ public sealed class MsBuildProps(AnalyzerOptions options)
     /// <summary>Indicates if the project is a test project.</summary>
     public bool? IsTestProject => Converts.String<bool>(Prop());
 
+    /// <summary>Indicates that CMP is enabled.</summary>
```
Frulfump: Nit: Typo
```suggestion
    /// <summary>Indicates that CPM is enabled.</summary>
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -1199,6 +1199,15 @@ public static partial class Rule
         tags: ["package", "import", "duplicate"],
         category: Category.Clarity);
 
+    public static DiagnosticDescriptor GlobalPackageReferencesOnlyWorkWithCpm => New(
+    id: 0813,
+    title: "Global package references only work when CPM is enabled",
+    message: "Global package reference '{0}' is ignored as CMP is disabled",
```
Frulfump: Nit: Typo
```suggestion
    message: "Global package reference '{0}' is ignored as CPM is disabled",
```

Frulfump (commented):
LGTM, found only nit typos.

AI Review (GPT 5.4) said this
> context.ManagePackageVersionsCentrally now prioritizes compilation-global analyzer config over the file graph. That is fine for the current project file, but risky for shared files like Directory.Packages.props, because they are analyzed in a project compilation context. A project-local opt-out can therefore bleed into shared-file analysis and cause false diagnostics. Consider reversing precedence for shared files, or using file/import discovery first and only falling back to compiler-visible props where file-local resolution is unavailable.

Corniel:
> AI Review (GPT 5.4) said this
> 
> > context.ManagePackageVersionsCentrally now prioritizes compilation-global analyzer config over the file graph. That is fine for the current project file, but risky for shared files like Directory.Packages.props, because they are analyzed in a project compilation context. A project-local opt-out can therefore bleed into shared-file analysis and cause false diagnostics. Consider reversing precedence for shared files, or using file/import discovery first and only falling back to compiler-visible props where file-local resolution is unavailable.

That is not true (completely). You get the actual resolved value while compiling that specific project. That also includes imported props/targets that were previously not taken into account. So the behavior might change, but in most cases for the better.

PR #746, opened 2026-04-20
Implement SemVer

Corniel:
To check if `<Version>` and `<VersionPrefix>` are compliant with [Semantic Versioning](https://semver.org/) we need a way to detect that first. I've opted for a simple implementation using a `Regex` that SemVer.org advised themselves.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/SemVer.cs
+++ b/src/DotNetProjectFile.Analyzers/SemVer.cs
@@ -1,0 +1,83 @@
+using System.Numerics;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace DotNetProjectFile;
+
+/// <summary>Implementation of Semantic Versioning.</summary>
+/// <remarks>
+/// See: https://semver.org/.
+/// </remarks>
+public sealed record SemVer()
+{
+    /// <summary>Initializes a new instance of the <see cref="SemVer"/> class.</summary>
+    public SemVer(
+        BigInteger major,
+        BigInteger minor,
+        BigInteger patch,
+        string? preRelease = null,
+        string? buildMetadata = null) : this()
+    {
+        Major = major;
+        Minor = minor;
+        Patch = patch;
+        PreRelease = preRelease;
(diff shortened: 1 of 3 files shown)


CptWesley:
Why not use the built-in `Version` type?

Corniel:
@CptWesley [System.Version](ttps://learn.microsoft.com/dotnet/api/system.version) does not follow SemVer as far as I know, or did I miss something?

src/DotNetProjectFile.Analyzers/SemVer.cs
```diff
@@ -0,0 +1,83 @@
...
+            Minor = BigInteger.Parse(match.Groups[nameof(Minor)].Value),
+            Patch = BigInteger.Parse(match.Groups[nameof(Patch)].Value),
+            PreRelease = match.Groups[nameof(PreRelease)].Value,
+            BuildMetadata = match.Groups[nameof(BuildMetadata)].Value,
+        },
+        _ => null,
+    };
+
+    private static readonly Regex Pattern = new(
+        @"^(?<Major>0|[1-9][0-9]*)\.(?<Minor>0|[1-9][0-9]*)\.(?<Patch>0|[1-9][0-9]*)(?:-(?<PreRelease>(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?<BuildMetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$",
```
Frulfump, line 80: Could really use a comment to describe this Regex and couldn't the source generator attribute be used instead here? https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators

Corniel: Nope. That is only not available for netstandard 2.0 unfortunately. :(

PR #750, opened 2026-04-21
Proj1002: Take actual MSBuild value into account

### LINKED ISSUE #747: Proj1002: Use Microsoft’s analyzers makes no sense on .NET 5+ projects
The rationale in the docs is
> For .NET 5 and later, is enabled by default. However, it is considered a good practice to explicitly enable it for later frameworks as well.

https://dotnet-project-file-analyzers.github.io/rules/Proj1002.html

But why? And by whom is it considered a good practice? It will just add noise since no later framework will ever change that default, for projects targeting `netstandard2.0` or `netstandard2.1` it could however make sense.

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-10#code-quality-analysis
> You can enable code analysis on projects that target earlier .NET versions by setting the [EnableNETAnalyzers](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enablenetanalyzers) property to true

And given https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/issues/652 where it's suggested to warn against setting the default value of a property.

Corniel:
By reading checking the MSBuild value, FP's can be prevented.

Closes #747

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseDotNetAnalyzers.cs
+++ b/src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseDotNetAnalyzers.cs
@@ -1,15 +1,19 @@
 namespace DotNetProjectFile.Analyzers.MsBuild;
 
+/// <summary>Impelements <see cref="Rule.UseDotNetAnalyzers"/>.</summary>
 [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
 public sealed class UseDotNetAnalyzers() : MsBuildProjectFileAnalyzer(Rule.UseDotNetAnalyzers)
 {
+    /// <inheritdoc />
+    public override bool DisableOnFailingImport => false;
+
     /// <inheritdoc />
     public override ImmutableArray<ProjectFileType> ApplicableTo => ProjectFileTypes.ProjectFile;
 
     /// <inheritdoc />
     protected override void Register(ProjectFileAnalysisContext context)
     {
-        if (!context.File.NETAnalyzersEnabled())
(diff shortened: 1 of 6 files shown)


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseDotNetAnalyzers.cs
```diff
@@ -1,15 +1,19 @@
 namespace DotNetProjectFile.Analyzers.MsBuild;
 
+/// <summary>Impelements <see cref="Rule.UseDotNetAnalyzers"/>.</summary>
 [DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
 public sealed class UseDotNetAnalyzers() : MsBuildProjectFileAnalyzer(Rule.UseDotNetAnalyzers)
 {
+    /// <inheritdoc />
+    public override bool DisableOnFailingImport => false;
```
Frulfump, line 8: What does this do?

Frulfump (approved):
Looks good to me, just one question

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/UseDotNetAnalyzers.cs
Corniel: Well, if we could not resolve any of the stated imports (most likely due to variables in a path we do not know the value of), we don't execute some rules, as we are not certain that we end up having the right value. As we ask MSBuild here what the truth is, we are certain about that, so don't care if we miss an import.

PR #770, opened 2026-04-23
Proj3000: Re-implement disk scan

### LINKED ISSUE #748: Seeing Proj3000: This file is using UTF-8 encoding with BOM for one of my dependencies
`<USER PATH>\.nuget\packages\microsoft.visualstudio.threading.analyzers\17.14.15\buildTransitive\AdditionalFiles\vs-threading.MainThreadSwitchingMethods.txt(1,1): warning Proj3000: This file is using UTF-8 encoding with BOM (https://dotnet-project-file-analyzers.github.io/rules/Proj3000.html)`

But that's out of my control to fix so why is there a warning for it?

Corniel:
By using a walker, some directories are not scanned at all (should make things faster): also, the filtering now also is applied on files that are included as additional files (FP's).

Closes #748 (again)


Sjaaky (approved):
lgtm

PR #780, opened 2026-04-28
Proj1001: Exclude private assets

### LINKED ISSUE #768: False positive Proj1001 xunit.analyzers to analyze xunit.runner.visualstudio
I don't think that analyzer package is intended to analyze xunit.runner.visualstudio
> warning Proj1001: Use xunit.analyzers to analyze xunit.runner.visualstudio (https://dotnet-project-file-analyzers.github.io/rules/Proj1001.html)

Using version 1.12.0

Corniel:
Closes #768

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Use_analyzers_for_packages.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Use_analyzers_for_packages.cs
@@ -133,7 +133,7 @@ public class Guards
           </PropertyGroup>
 
           <ItemGroup>
-            <PackageReference PrivateAssets="all" Include="xunit.runner.visualstudio" Version="3.1.5" />
+            <PackageReference Include="xunit.runner.visualstudio" Version="3.1.5" />
           </ItemGroup>
 
         </Project>
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Use_analyzers_for_packages.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Use_analyzers_for_packages.cs
@@ -125,7 +125,7 @@ public class Guards
        .HasNoIssues();
 
     [Test]
-    public void xunit_runner_visualstuidio_as_private_asset() => new UseAnalyzersForPackages().ForInlineCsproj("""
+    public void xunit_runner_visualstudio_as_private_asset() => new UseAnalyzersForPackages().ForInlineCsproj("""
         <Project Sdk="Microsoft.NET.Sdk">
 
           <PropertyGroup>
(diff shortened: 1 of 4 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Use_analyzers_for_packages.cs
```diff
@@ -124,6 +124,22 @@ public class Guards
         """)
        .HasNoIssues();
 
+    [Test]
+    public void xunit_runner_visualstuidio_as_private_asset() => new UseAnalyzersForPackages().ForInlineCsproj("""
```
Frulfump: Typo 
```suggestion
    public void xunit_runner_visualstudio_as_private_asset() => new UseAnalyzersForPackages().ForInlineCsproj("""
```

Frulfump (changes requested):
I’m not convinced `PrivateAssets="all"` is sufficient as a general exclusion criterion. A private package can still be a package whose code/API usage should be analyzed; `PrivateAssets` only controls transitive asset flow to consuming projects.

For the current false positive, a narrower exclusion may be safer: exclude `xunit.runner.visualstudio` from the `xunit` analyzer match, or introduce an explicit “do not analyze these package IDs” list for known adapter/build/analyzer packages. The current rule already has package-specific analyzer matching, so a package-specific exclusion would avoid creating false negatives for private packages that still should be analyzed.

If you want to keep the generalization, I’d at least constrain it to `PrivateAssets="all"` and add tests proving that partial `PrivateAssets` values such as `contentFiles` or `contentFiles;analyzers` are still analyzed.

Currently it's accepting any value in `PrivateAssets` not just `all`

Corniel:
> I’m not convinced `PrivateAssets="all"` is sufficient (..)
>
> For the current false positive, a narrower exclusion may be safer: exclude `xunit.runner.visualstudio` (..)

I guess you're right: I've updated the implementation.

specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Use_analyzers_for_packages.cs
```diff
@@ -124,6 +124,22 @@ public class Guards
...
+            <TargetFramework>net10.0</TargetFramework>
+          </PropertyGroup>
+
+          <ItemGroup>
+            <PackageReference Include="xunit.runner.visualstudio" Version="3.1.5" />
+          </ItemGroup>
+
+        </Project>
+        """)
+      .HasNoIssues();
```
Frulfump, line 141: Just thought about this one now, should it check for the specific issue to minimize churn if new rules are added/updated?

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -83,6 +83,7 @@ v.13.0
 - Proj3000: Drop difference between additional and regular files. (FP)
 v1.12.2
 - Proj0201: Exclude Microsoft.NET.Sdk.Web projects. (FP)
+- Proj1001: Exclude private assets. (FP)
```
Frulfump: This should be modified then I think

Frulfump (approved):
LGTM

Issue #756, opened 2026-04-22
Proj0044 triggered even when `ContinuousIntegrationBuild` is defined with a condition in `Directory.Build.props` to only be included in CI


Frulfump:
I have this in a `Directory.Build.props` in my root   

```xml
...
<ContinuousIntegrationBuild Condition="'$(GITHUB_ACTIONS)' == 'true'">true</ContinuousIntegrationBuild>
<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
...
```

But I still get Proj0044 reported for a project file when doing a local build (not in GitHub Actions)
https://dotnet-project-file-analyzers.github.io/rules/Proj0044.html

The condition means the property isn't included when it's not true

Corniel:
You miss the `<RestoreLockedMode>` property?!

Frulfump:
> You miss the `<RestoreLockedMode>` property?!

I run with locked mode in CI with the command line option. I think this FP is triggered due to the condition not being taken into account?
UPDATE: I have this in my `Directory.Build.props` as well `<RestoreLockedMode Condition="'$(GITHUB_ACTIONS)' == 'true'">true</RestoreLockedMode>`

Corniel:
With `--locked-mode` you mean? I never tested if a rule could check that, I assume it does, and that the value is known in MSBuild.

Corniel:
@Frulfump can you review the changes?

Frulfump:
> With `--locked-mode` you mean? I never tested if a rule could check that, I assume it does, and that the value is known in MSBuild.

Yes exactly!

Frulfump:
This seems to still happen with 1.12.0
>warning Proj0044: Define the <RestoreLockedMode> node with value 'true' when <ContinuousIntegrationBuild> is enabled (https://dotnet-project-file-analyzers.github.io/rules/Proj0044.html)

` <ContinuousIntegrationBuild>` isn't included as there's a condition for it to only be included and set to `true` when I'm running in GitHub Actions but I still see it, so maybe the detection code is ignoring that the property is conditioned?

Corniel:
I'll look into this.

Frulfump:
> I'll look into this.

Thanks! I updated the title to better reflect the actual issue.
[...]

Issue #732, opened 2026-04-17
Use MSBuild variables to include Directory.Build.* files


Corniel:
Use `ImportDirectoryBuildProps` and `DirectoryBuildPropsPath` to import `Directory.Build.*` files. This should also ensure we do not longer need direct disk asccess to resolve these.

Corniel:
This has been solved with #753

Issue #483, opened 2025-06-30
Use Warpstone for non-XML based files


Corniel:
The use of `Antlr` (with all its dependencies) nor the homegrown `Grammr` resulted in a solution I'm happy with. So lets try @CptWesley 's [Warpstone](https://github.com/CptWesley/Warpstone).

Corniel:
@CptWesley Any time to pick this up?

Corniel:
We've went with Grammr, which works okay.

Issue #33, opened 2023-07-09
Execute condtions


Corniel:
nodes can be defined conditionally. Those conditions should be taken into account. 

See:
https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-conditions

Corniel:
I'm closing this one. For resolving paths we need a mechanism (which is addressed) but for other conditions we should consider all paths, as those paths are there for a reason (mostly), as they are all possible.

PR #782, opened 2026-04-28
Proj0033: MSBuild property substitution

### LINKED ISSUE #546: Proj0022: Resolves paths from incorrect directory when in `Directory.Build.props`
When paths are resolved from `Directory.Build.props` the base directory is the one where the `.csproj` is located, not where the `Directory.Build.props`. It appears that for Proj0022 (and presumably also for other rules) this causes both false positives and false negatives.

### LINKED ISSUE #778: Proj0033 false positive: $(MSBuildThisFileDirectory) (and other MSBuild properties) not expanded before file-existence check
`Lib/Lib.csproj`:

```xml
  <PropertyGroup>
  </PropertyGroup>
  <ItemGroup>
    <ProjectReference Include="$(MSBuildThisFileDirectory)..\My.Analyzer\My.Analyzer.csproj"
                      PrivateAssets="all"
                      ReferenceOutputAssembly="false"
                      OutputItemType="Analyzer" />
  </ItemGroup>
</Project>
```

`My.Analyzer/My.Analyzer.csproj` exists and the `Include` resolves correctly at build time
(MSBuild expands `$(MSBuildThisFileDirectory)` to the absolute path of the directory holding the
`.csproj`). The project builds, links, and runs. Yet:

```
Lib\Lib.csproj(6,5): warning Proj0033: The Include '$(MSBuildThisFileDirectory)..\My.Analyzer\My.Analyzer.csproj' of <ProjectReference> does not exist
```

The error message quotes the unexpanded property literally, suggesting the existence check runs
against the raw XML attribute value rather than the MSBuild-evaluated path. References that use
plain relative paths (`..\My.Analyzer\My.Analyzer.csproj`) do not fire the rule.

The same blind-spot also makes the rule fire on `<ProjectReference>` items inside an
`<ItemGroup Condition="…">` even when the Condition would exclude the item — likely the same
"don't evaluate, just check the literal text" code path.

Closest open precedent: #756 (Proj0044 with a Condition) and the closed-fixed sibling fixes in
#258 (Proj0009) and #257 (Proj0028).

Workaround: `<NoWarn>$(NoWarn);Proj0033</NoWarn>`.

JohnVerheij:
## Summary

- Closes #778 (MSBuild property `$(MSBuildThisFileDirectory)` and friends now expanded before the file-existence check on `Proj0033`/`Proj0022`)
- Closes #546 (relative item paths declared in `Directory.Build.props` resolve against the consuming `.csproj`, plus the same canonicalisation now applies to `IncludeProjectReferencesOnce`)

The condition-evaluation work that addresses Issue #33 is deferred to a follow-up after the Warpstone parser migration discussed at #483 lands and a plan for #33 is agreed.

## What this PR adds

1. **`PropertyRegistry`.** Resolves `$(...)` references in path expressions. 12 reserved per-file / per-project properties (`MSBuildThisFile*`, `MSBuildProject*`, including the `*NoRoot` variants). User-defined properties resolve via an unconditional `<PropertyGroup>` scan; conditional groups are skipped (the deferred condition work).
2. **`Resolve` on `ProjectFileAnalysisContext`.** Returns `(IODirectory Root, string Literal)`. Absolute substituted paths split into root + filename; relative paths and inputs with unresolved properties anchor at `context.File.Path.Directory` (the consuming project), which is the documented MSBuild rule for items declared in imports.
3. **`Substitute` and `Files` extensions.** Direct access to substitution output and a substitution-aware `Files(node, include)` for rules that previously called `root.Files(include)` against an unsubstituted include.
4. **`IncludeProjectReferencesOnce` cross-file canonicalisation.** Duplicate-detection key is now built through `Resolve`, so a reference in the `.csproj` and a textually equivalent one in `Directory.Build.props` (`$(MSBuildProjectDirectory)/...`) canonicalise to the same `IOFile` and the duplicate fires.
5. **`Proj0052` informational diagnostic.** `MsBuildPropertyCouldNotBeResolved` surfaces `$(...)` references the registry cannot resolve. Severity `Info`, `isEnabled: false` by default; users opt in via `.editorconfig` when they want to be told about unregistered references.
6. **`MSBuildProjectDirectory` `<CompilerVisibleProperty>`** registration plus matching `MsBuildProps` accessor, to surface the value through the `build_property` pipeline alongside the 16 properties already registered.

## Bug fixes

**Malformed `$(MSBuildProjectDirectory)Foo.csproj` is now flagged.** The previous `Resolve` matched the literal string and stripped it, silently accepting input that real MSBuild treats as `<dir>Foo.csproj` and would fail at build time. The substitution layer follows MSBuild's literal-expansion semantics, so the malformed form now correctly fires `Proj0033`. Two PR-internal test fixtures were corrected to `$(MSBuildProjectDirectory)/Foo.csproj`.

**Relative item paths in `Directory.Build.props` resolve against the consuming project (#546).** MSBuild's documented rule is that relative item paths evaluate against `$(MSBuildProjectDirectory)` regardless of which imported file declares the item. `Resolve` previously anchored the bare-relative fallback to the containing file's directory, producing false positives and false negatives for any rule walking items in imports. `Resolve` now anchors at `context.File.Path.Directory`.

**`IncludeProjectReferencesOnce` cross-file duplicate detection fires (#546).** The duplicate-detection key was built from `r.Project.Path.Directory.File(r.Include)` without property substitution, so a `<ProjectReference>` declared in the `.csproj` and a textually equivalent one in `Directory.Build.props` canonicalised to different paths and the duplicate was silently missed. The rule now routes the include through `context.Resolve(...)`.

## Tests

- 55 tests on `PropertyRegistry` (all 12 reserved properties, case insensitivity, empty input, malformed references, recursion guard, anchor distinction between containing and entry-point file, all 4 directory-format variants).
- 8 tests on `MsBuildPropertyShouldBeResolvable` (reserved skip, user-defined unconditional resolve, conditional `<PropertyGroup>` skip, conditional child skip, single unresolved on `<ProjectReference>` and on `<None>`, multiple distinct unresolved each fire, literal include).
- 5 integration tests in `Cross_file_path_resolution.cs` covering the dual-anchor distinction (#546 surface): bare path resolves against the consuming `.csproj` (positive and negative), `$(MSBuildThisFileDirectory)` resolves against the props file, `$(MSBuildProjectDirectory)` resolves against the entry-point `.csproj`, `$(MSBuildThisFileDirectory)` inside `Directory.Build.targets` resolves against the targets file.
- Per-rule tests added across `Project_reference_include_should_exist`, `Project_reference_includes_should_be_case_consistant`, `Build_action_include_should_exist`, `Include_project_references_once` covering both the property-substitution surface and the cross-file duplicate detection.

All 3,685 tests in the test suite pass except the pre-existing Proj1001 transitive-NuGet flake. No existing tests were modified.

## Out of scope

- **Condition evaluation (#33).** Deferred. Conditional `<ItemGroup Condition="...">` is still walked unconditionally; conditional `<PropertyGroup Condition="...">` is now skipped during the user-defined-property scan. Adding a real evaluator depends on the Warpstone migration discussed at #483 plus an agreed plan for #33; the follow-up will land after both.
- **#756 (Proj0044 false positive on conditionally-defined `ContinuousIntegrationBuild`).** Same dependency as #33.
- **#732 (use `DirectoryBuildPropsPath` CVP for `Directory.Build.*` resolution).** The `MSBuildProjectDirectory` CVP work makes this reachable, but the actual fix requires plumbing the analysis context through `MsBuildProject.DirectoryBuildProps` access. Distinct concern.
- **`Microsoft.Build` / Buildalyzer dependency for full evaluation.** `PropertyRegistry` uses no external deps; reserved properties are derived from `IOFile` paths, user-defined properties come from a direct `<PropertyGroup>` scan.
- **Numeric / boolean comparisons, negation, functions other than `Exists`.** All deferred with the condition evaluator.

## Reading order

1. `PropertyRegistry` (the core type)
2. `Resolve` rewrite on `ProjectFileAnalysisContext` (the integration point)
3. `IncludeProjectReferencesOnce` (the cross-file canonicalisation fix)
4. `BuildActionIncludeShouldExist` / `ProjectReferenceChecker` / `ProvideCompliantPackageIcon` (the rules using the new `Resolve` + `Files` + `Substitute`)
5. `MsBuildPropertyShouldBeResolvable` (Proj0052)
6. The integration tests (`Cross_file_path_resolution.cs`)


CptWesley:
I'm wondering if the warning message should include the resolved path instead/in addition to the current path. It would maybe make it easier for someone to find out what _exact_ path is expected and is missing.

src/DotNetProjectFile.Analyzers/Extensions/DotNetProjectFile.Diagnostics.ProjectFileAnalysisContext.cs
```diff
@@ -15,5 +18,33 @@ public bool EnableNETAnalyzers
...
+        /// to the search-root directory and the literal portion of the include.
+        /// </summary>
+        /// <remarks>
+        /// Supports <c>$(MSBuildThisFileDirectory)</c> (the directory of the file containing
+        /// <paramref name="node"/>) and <c>$(MSBuildProjectDirectory)</c> (the directory of the
+        /// entry-point project file). Includes containing other unresolved properties are returned
+        /// unchanged, falling through to a literal lookup against the node's project directory.
+        /// </remarks>
+        public (IODirectory Root, string Literal) Resolve(Node node, string include)
+            => (include.Contains(MsBuildThisFileDirectory), include.Contains(MsBuildProjectDirectory)) switch
```
CptWesley: Won't this result in strange behaviour if these environment variables are not the segment of the path? I can imagine this being a very strange and unlikely scenario. Ideally, I would prefer to support this fully. But if we only want to cover the case of it being the first segment, I would prefer this being a `.StartsWith`, rather than `.Contains`. But I will let that up to @Corniel to decide

src/DotNetProjectFile.Analyzers/Extensions/DotNetProjectFile.Diagnostics.ProjectFileAnalysisContext.cs
```diff
@@ -15,5 +18,33 @@ public bool EnableNETAnalyzers
...
+            {
+                (true, false) => (node.Project.Path.Directory, include.Replace(MsBuildThisFileDirectory, string.Empty)),
+                (false, true) => (context.File.Path.Directory, include.Replace(MsBuildProjectDirectory, string.Empty)),
+                _ => (node.Project.Path.Directory, include),
```
CptWesley: If we are opting for full support, I would also expect a `(true, true)` case.

JohnVerheij:
@CptWesley on the warning message:

Good suggestion. I'd lean toward showing both, with the literal include for
source orientation and the resolved path showing the actual missing file.
Happy to do replacement-only if you'd prefer. Will fold this in once the
resolution direction is settled in the other thread, since the
resolved-path output depends on which approach lands.

src/DotNetProjectFile.Analyzers/Extensions/DotNetProjectFile.Diagnostics.ProjectFileAnalysisContext.cs
JohnVerheij: Thanks for the review. This PR is scoped to #778, which reports the false
positive for the start-of-path case (`$(MSBuildThisFileDirectory)Foo.cs`).
`.StartsWith` matches that scope cleanly. Mid-string and double-property
handling are legitimate concerns but expand beyond what was reported. The
only MSBuild-property handling in the repo today is the
`IODirectory.Iterate` bail-out on `$(` (line 180, comment "We do not
support variables yet") plus literal-string matching of `Condition`
attributes via `HasAnyCondition`. No expression evaluator exists, and
there are zero `Microsoft.Build.*` references in the codebase. Happy to
file a follow-up issue for full property substitution where the design
space (path normalization, drive-letter handling, scope of supported
properties) can be discussed properly. For this PR, lean toward
`.StartsWith` to ship the reported fix. Open to either direction once
you and @Corniel decide.

src/DotNetProjectFile.Analyzers/Extensions/DotNetProjectFile.Diagnostics.ProjectFileAnalysisContext.cs
JohnVerheij: Moot if we land on `.StartsWith` (two directory-prefix properties can't
both be at the start). If full support is preferred, the (true, true)
case would be handled alongside the broader resolver work, which I've
suggested in the other thread as a follow-up issue rather than expanding
this PR.

Corniel:
I would expect `MsBuildProps` to be extended with these properties of the type `IODirectory`. For `$(MSBuildThisFileDirectory)` it might/should be resolvable based on the current file/path.

I think we should also aim to replace `\$\(?<Property>\w+\)`. What to do if the regex does not match anything is another question. As suggested: the two core props should only be supported when started with it.

JohnVerheij:
For `MSBuildProjectDirectory` on `MsBuildProps`: `MSBuildProjectFile` returns the filename only, so the directory has to come from elsewhere. Either extend the `MsBuildProps` ctor to accept `IOFile? projectFile` and surface `MSBuildProjectDirectory => projectFile?.Directory ?? IODirectory.Empty`, or add `<CompilerVisibleProperty Include="MSBuildProjectDirectory" />` to `DotNetProjectFile.Analyzers.Sdk.props` and read it via `Prop()`. Which do you prefer?

`MSBuildThisFileDirectory` does not fit `MsBuildProps`. Its value depends on the node being analysed (importing `.props` vs entry-point `.csproj`), so it cannot be a parameterless property on a context-level instance. I'd resolve it in the substitution layer via `node.Project.Path.Directory`.

The regex: assuming `\$\((?<Property>\w+)\)`. Correct?

For the unresolved-property case, three options: leave the literal `$(Foo)` in the path and let the file-existence check skip silently; suppress the diagnostic entirely (matching the `$(` bail-out in `IODirectory.Iterate`); or emit a separate "unrecognised MSBuild property" diagnostic. Which?

`.StartsWith` on the two directory props stays as you and @CptWesley settled.

Corniel:
I'm not sure if the regex is correct. I did not test that (yet).

In general: the file walker that is now not aware of the MSBuild/Roslyn context should become aware of it. You're going in the right direction, and we might want to do it in multiple steps. If It wasn't this challenging, I would already have done it I guess. ;)

JohnVerheij:
Expanded the PR per the architectural redirect. PropertyRegistry + ConditionEvaluator + condition-aware iteration helpers are now in place; closes #778, #33, #546. Rebased onto current upstream/main, with conflicts resolved against #753's SDK refactor. See the updated description for capabilities, rule migrations, bug fixes, tests, and out-of-scope items.

Corniel:
I think that I would prefer to keep (conditional) location resolving separated from conditional analysis, as they are different beasts. The latter covers #33 and has a lot of problems on it own. Most notoriously that we only have limited access to the state of MSBuild, and that can not be overcome easily.

Furthermore, your using Grammr, an homegrown parser. Although I like the attempt myself, @CptWesley and discussed the idea to move to Warpstone  #483 (made by @CptWesley ). So, please, lets us first ensure that we can find more file/directory references and once that is shipped, together come with a plan to solve #33 (not to be confused with Proj0033).

JohnVerheij:
Understood. I'll reduce this PR to the path/property-resolution scope: PropertyRegistry, Resolve refactor (entry-point-anchored fallback fixes #546), IncludeProjectReferencesOnce cross-file canonicalisation fix, MSBuildProjectDirectory CVP + accessor, Proj0052. Closes #778 and #546.

The condition-evaluation work (ConditionEvaluator on Grammr, IsConditionEnabled, `EnabledItems<T>`, the 15-rule migration, ScanUserDefinedProperty's PropertyGroup condition logic) comes out and waits on the Warpstone landing plus a #33 plan.

I'll push the reduced version when ready.

JohnVerheij:
Reduced version pushed. Now scoped to path/property resolution per the redirect; closes #778 and #546. The condition-evaluation work waits on Warpstone per #483 and a #33 plan.

Corniel:
I'm having doubts about proposed rule Proj0052. From a developer (of this project) perspective it might be great to have it, but from the perspective of a user (someone who just want to get insights in its build files) it is not. So even though the proposed config (info, disabled by default) is still is suggested that people might want to enable it, and as a result make some properties `<CompilerVisible>`. I think a rule like that should be in the 9000 range, indicating to be DEBUG only rules.

Another problem we have to address (this PR made me think of the this) is that some conditions in there, might be resolvable for the current build/MSBuild values, but should be evaluated agnosticly,  because they where put in there to cover multiple scenarios (UNIX v.s. Windows, multiple target frameworks, local v.s. build enviroments, etc.)

So by ignoring conditions that we can resolve as not being active, we might miss something.

Another challenge might lay in scenarios where we want to check for project X that depends on project Y. If we have to evaluate project Y to say something about, lets say `<PackageReference>`s, than we should be able to evaluate Y, with the assumed context for Y, and not X. This is not trivial at all.

JohnVerheij:
Proj0052 → 9000 range works. Hidden by default makes more sense than Info, given the rule is analyzer-internal introspection rather than something users might enable. The description could drop the "register or refactor" nudge in the same change; that's the bit that risks pushing users toward CompilerVisibleProperty edits, which you're right to avoid.

The condition-agnostic point reframes the property scanner from "skip conditional declarations" to "surface every declaration with its gating condition expression preserved as metadata." Without a real condition evaluator, the conservative move is to preserve the information rather than drop it; rules then pick a policy (any-branch, all-branches, last-unconditional, per-TFM). That's a wider change than Proj0052 but benefits every rule that touches PropertyGroup elements. Doable on top of the current scanner without breaking existing call sites if the new entry point is additive and the existing one becomes a PickUnconditional wrapper over it.

Cross-project context (X depending on Y, evaluate Y in Y's context) is the foundational one. The "obvious" answer would be Microsoft.Build.Graph.ProjectGraph, but it's not safe to reference from a Roslyn analyzer. Assembly-load conflicts in devenv.exe are a recurring category (open [dotnet/roslyn#41421](https://github.com/dotnet/roslyn/issues/41421)), MSBuildLocator timing constraints, and graph construction sits several orders of magnitude over the per-keystroke analyzer budget. The right answer looks more like an XDocument-based ProjectReference walker that loads each referenced project in its own evaluation context (not the entry project's), with conditions still preserved as expression metadata per the second point. Cycle-tolerant, depth-bounded, reuses the existing ProjectFiles cache as the loader.

One thing your second and third points together suggest: making multi-TFM a per-rule declared stance. TFM-agnostic by default (current behavior), with opt-ins for "fire under any TFM," "fire under every TFM," and "fire per TFM." Sits naturally between conditions-as-metadata and per-project-context-isolation, and lets each rule say explicitly what its multi-target story is.

No urgency on my side, happy to keep iterating as long as useful. If the cross-project context feels better as a separate design issue, that works too.

JohnVerheij:
Pushed ec915ae addressing the review feedback:

- Proj0052 -> Proj9000 (debug range, Hidden, disabled by default).
  The rule no longer appears in user-facing docs and is filtered
  from the globalconfig snapshot by the existing >=9000 convention
  in `Design_specs`.
- Rule description rewritten without the "register or refactor"
  wording.
- Resolved path surfaces in Proj0033 and Proj0022 messages when
  the include uses MSBuild property substitution, as a project-
  relative path so test assertions stay deterministic. Per
  CptWesley's earlier request.
- `PropertyRegistry`'s user-defined property scan now skips
  `<PropertyGroup>` nested inside `<Target>`, matching the
  exclusion #801 added for the typed property lookup. New spec
  test covers it.
- `IODirectory`'s file walker now guards against directories that
  do not exist after MSBuild property substitution. The crash
  surfaced while running a Proj0033 spec test against
  `$(MSBuildThisFileDirectory)Sub/...` where `Sub/` did not exist
  on disk; the analyzer now reports Proj0033 cleanly instead of
  throwing `DirectoryNotFoundException`.
- A Sonar S125 false positive on the doc comment of our
  integration spec was cleaned up.

JohnVerheij:
Closing this one. It was opened in April and has been untouched since mid-May, and `main` has moved a long way underneath it; with the `AnalyzerType` work, the lexer changes and the SDK pin all touching the same files, rebasing five commits across 27 files costs more than re-deriving the parts that still matter.

Worth flagging so they do not get lost: **#778 and #546 are still open.** Those were the two bugs this fixed, and neither depends on the condition-evaluation work you redirected away from.

The core of it was `PropertyRegistry` plus the entry-point-anchored `Resolve` refactor. #954 sits close to that: `IOPath.IsFullyQualified` reads the literal text, so an `<OutputRoot>C:\Builds</OutputRoot>` followed by `<OutputPath>$(OutputRoot)\bin\</OutputPath>` is flagged on neither line. Same gap, if the property side is ever worth picking up on its own.

Thanks for the reviews back in May. Keeping location resolution separate from condition evaluation was the right call.

PR #783, opened 2026-04-28
Proj0253/4: Fix runtime TypeConverter resolvement

### LINKED ISSUE #773: Proj0253: Version '2.3.2' does not comply with Semantic Versioning
I've got this strange warning on the [Qowaiv.Analyzers project](https://github.com/Qowaiv/qowaiv-analyzers/blob/main/src/Qowaiv.CodeAnalysis.CSharp/Qowaiv.CodeAnalysis.CSharp.csproj). Strange as the tests in this project seem to work, and the value `2.3.2` is recognized - apparently - by the Proj0253 rule. I also tryed `2.4.0-rc1` with the same result.

### LINKED ISSUE #779: Proj0254 false positive: VersionPrefix '1.0.0' (and any other plain X.Y.Z value) reported as not SemVer-compliant
`Directory.Build.props`:

```xml
<Project>
  <PropertyGroup>
    <VersionPrefix>1.0.0</VersionPrefix>
  </PropertyGroup>
</Project>
```

Any child project (verified on `net10.0` and `netstandard2.0`):

```
Directory.Build.props(N,M): warning Proj0254: VersionPrefix '1.0.0' does not comply with Semantic Versioning
```

`1.0.0` is the canonical valid SemVer 2.0 example. The same error fires for `0.23.0`, `2.5.7`,
and every other `X.Y.Z` value tested. The rule's documented "non-compliant" example
(`3.1.41596-beta0.03`) carries a `-prerelease` suffix that the rule legitimately wants out of
`VersionPrefix` and into `VersionSuffix`, but the rule's actual implementation also rejects
plain `MAJOR.MINOR.PATCH` values — i.e. essentially every non-prerelease version anyone would
ever set on a stable package.

Source-level check (`VersionPrefixShouldBeSemVerCompliant.cs`):

```csharp
private static bool NotSemantic(VersionPrefix version)
    => version.Value is not { PreRelease: null, BuildMetadata: null };
```

For `1.0.0`, `PreRelease` and `BuildMetadata` are both null and the underlying `SemVer.TryParse`
should succeed → `NotSemantic` should be `false`. Empirically it is `true`, so either
`SemVer.TryParse` is returning `null` for valid input or the property pattern is being inverted
somewhere upstream.

Sibling rule already reported by the maintainer: #773 (Proj0253 false positive on `2.3.2`).
Originating PR: #749 (shipped in 1.12.0).

Workaround: `<NoWarn>$(NoWarn);Proj0254</NoWarn>`.

Corniel:
It turns out that the `TypeConverter` resolvement does work in our test setup, but not within a `dotnet build` context. By simplifying this dramatically ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)) the problem was resolved.

Closes #773
Closes #779

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Semantic_versioning_specs.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Semantic_versioning_specs.cs
@@ -41,6 +41,7 @@ public class Reports
 
 public class Guards
 {
+    [TestCase("1.0.0")]
     [TestCase("3.14.1596")]
     [TestCase("1.1.0-rc.1+e471d15")]
     public void semantic_version(string version) => new VersionShouldBeSemVerCompliant()
--- a/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
@@ -8,12 +8,11 @@ public static class TypeConverters
 {
     public static T? TryConvert<T>(string? value)
     {
-        var type = Nullable.GetUnderlyingType(typeof(T)) ?? typeof(T);
         try
         {
             if (value is { Length: > 0 })
             {
-                return (T?)Get(type).ConvertFromInvariantString(value);
+                return (T?)Get(typeof(T)).ConvertFromInvariantString(value);
             }
         }
         catch
@@ -24,9 +23,12 @@ public static class TypeConverters
     }
 
     public static TypeConverter Get(Type type)
-        => TypeStore.TryGetValue(type, out var custom)
-        ? custom
-        : TypeDescriptor.GetConverter(type);
+    {
+        var not_null = Nullable.GetUnderlyingType(type) ?? type;
+        return TypeStore.TryGetValue(not_null, out var custom)
+            ? custom
+            : TypeDescriptor.GetConverter(not_null);
+    }
 
     private static readonly FrozenDictionary<Type, TypeConverter> TypeStore = new Dictionary<Type, TypeConverter>()
     {
--- a/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
@@ -4,19 +4,15 @@ using System.ComponentModel;
 
 namespace DotNetProjectFile.MsBuild.Conversion;
 
-internal sealed class TypeConverters
+public static class TypeConverters
 {
-    public T? TryConvert<T>(string? value)
+    public static T? TryConvert<T>(string? value)
     {
         try
         {
             if (value is { Length: > 0 })
             {
-                var converter = TypeStore.TryGetValue(typeof(T), out var custom)
-                    ? custom
-                    : TypeDescriptor.GetConverter(typeof(T));
-
-                return (T?)converter.ConvertFromInvariantString(value);
+                return (T?)Get(typeof(T)).ConvertFromInvariantString(value);
             }
         }
         catch
@@ -26,7 +22,12 @@ internal sealed class TypeConverters
         return default;
     }
 
-    private readonly FrozenDictionary<Type, TypeConverter> TypeStore = new Dictionary<Type, TypeConverter>()
+    public static TypeConverter Get(Type type)
+        => TypeStore.TryGetValue(type, out var custom)
+        ? custom
+        : TypeDescriptor.GetConverter(type);
+
+    private static readonly FrozenDictionary<Type, TypeConverter> TypeStore = new Dictionary<Type, TypeConverter>()
     {
         [typeof(string)] = new StringConverter(),
         [typeof(bool)] = new BooleanConverter(),
--- a/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
@@ -8,11 +8,12 @@ public static class TypeConverters
 {
     public static T? TryConvert<T>(string? value)
     {
+        var type = Nullable.GetUnderlyingType(typeof(T)) ?? typeof(T);
         try
         {
             if (value is { Length: > 0 })
             {
-                return (T?)Get(typeof(T)).ConvertFromInvariantString(value);
+                return (T?)Get(type).ConvertFromInvariantString(value);
             }
         }
         catch
--- a/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
+++ b/src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
@@ -1,18 +1,22 @@
 using DotNetProjectFile.Conversion;
+using System.Collections.Frozen;
 using System.ComponentModel;
-using System.Reflection;
 
 namespace DotNetProjectFile.MsBuild.Conversion;
 
 internal sealed class TypeConverters
 {
-    public T? TryConvert<T>(string? value, Type classType, string propertyName)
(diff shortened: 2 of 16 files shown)


specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Semantic_versioning_specs.cs
JohnVerheij, line 62: The new `[TestCase("1.0.0")]` is on `Guards.semantic_version` (line 47) which exercises `VersionShouldBeSemVerCompliant` (Proj0253 / `<Version>`). The reported issue #779 was on Proj0254 (`<VersionPrefix>`). Could parametrize this test the same way to add `[TestCase("1.0.0")]` directly here. That would close the loop on #779's reported case explicitly. The parser is shared so the fix is covered transitively, but explicit coverage matches the reported bug verbatim.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
```diff
@@ -1,18 +1,22 @@
...
-    public T? TryConvert<T>(string? value, Type classType, string propertyName)
+    public T? TryConvert<T>(string? value)
     {
         try
         {
-            if (value is { })
+            if (value is { Length: > 0 })
             {
-                return (T?)Get(classType, propertyName!).ConvertFromInvariantString(value);
+                var converter = TypeStore.TryGetValue(typeof(T), out var custom)
```
JohnVerheij: For value types (`bool`, `IOFile`, `IODirectory`, `LanguageVersion`), this lookup misses. `Node<T>.Value` calls `Convert<T?>`, so `typeof(T)` here is the nullable form (e.g., `typeof(bool?)` = `typeof(Nullable<bool>)`), while the `TypeStore` keys are the non-nullable types. Verified locally:

- `typeof(bool) != typeof(bool?)` (true)
- `typeof(IOFile) != typeof(IOFile?)` (true)
- `typeof(IODirectory) != typeof(IODirectory?)` (true)
- `TypeDescriptor.GetConverter(typeof(bool?))` returns `NullableConverter`, not the `BooleanConverter` from `TypeStore`.

For `SemVer` (record/reference type), the C# compiler doesn't allow `typeof(SemVer?)` at all (nullable annotation is metadata-only), so `typeof(T)` collapses to `typeof(SemVer)` and the entry IS hit.

Functionally equivalent (the `NullableConverter` fall-through wraps the type-level `[TypeConverter]` attribute correctly), but the new `IOFile` / `IODirectory` entries don't actively contribute via this path. Could be made direct by unwrapping nullable before the lookup:

```csharp
var lookupType = Nullable.GetUnderlyingType(typeof(T)) ?? typeof(T);
var converter = TypeStore.TryGetValue(lookupType, out var custom)
    ? custom
    : TypeDescriptor.GetConverter(typeof(T));

File: src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs [hunk 2]
```diff
@@ -1,18 +1,22 @@
...
 
 internal sealed class TypeConverters
 {
-    public T? TryConvert<T>(string? value, Type classType, string propertyName)
+    public T? TryConvert<T>(string? value)
     {
         try
         {
-            if (value is { })
+            if (value is { Length: > 0 })
```
JohnVerheij, line 13: The guard moves from `value is { }` to `value is { Length: > 0 }`, so empty strings now skip the converter and return `default(T)`. For most converters this is a no-op (they throw on empty, caught, return default), but a few previously returned non-default values for empty input via the convert path. Verified locally:

- `new StringConverter().ConvertFromInvariantString("")` returns `""` (not null)
- `IOFile.Parse("")` returns `IOFile.Empty` (not null)
- `IODirectory.Parse("")` returns `IODirectory.Empty` (not null)

So `Convert<string>("")` previously returned `""`, now returns `null`; `Convert<IOFile?>("")` previously returned `IOFile.Empty`, now returns `null`. Worth a quick sanity check that nothing downstream distinguishes "explicitly empty element" from "missing element". If the behavior change is intentional, it could also be a separable commit from the SemVer fix (different concern, easier to reason about).

JohnVerheij (approved):
Looks good. Clean simplification. Replacing the reflection-based property-attribute lookup with a `FrozenDictionary<Type, TypeConverter>` first-check + `TypeDescriptor.GetConverter` fallback removes both the locking and the property cache. Tests cover both #773 and #779 directly via the new `[TestCase("1.0.0")]` and the existing `semantic_version_prefix` Guards test.

Three non-blocking observations inline.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
```diff
@@ -22,44 +26,14 @@ internal sealed class TypeConverters
...
         [typeof(string)] = new StringConverter(),
         [typeof(bool)] = new BooleanConverter(),
+        [typeof(IOFile)] = new IOFileConverter(),
+        [typeof(IODirectory)] = new IODirectoryConverter(),
         [typeof(LanguageVersion)] = new LanguageVersionConverter(),
-    };
-
-    private readonly object locker = new();
+        [typeof(SemVer)] = new SemVerConverter(),
+    }
```
CptWesley, line 41: Perhaps also include a design test case that scans the assembly for all implemented type converters and checks if they are registered here correctly. Just to make sure we don't forget that.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
```diff
@@ -1,18 +1,22 @@
...
+using System.Collections.Frozen;
 using System.ComponentModel;
-using System.Reflection;
 
 namespace DotNetProjectFile.MsBuild.Conversion;
 
 internal sealed class TypeConverters
 {
-    public T? TryConvert<T>(string? value, Type classType, string propertyName)
+    public T? TryConvert<T>(string? value)
```
CptWesley: Perhaps we need some unit tests for this method for the various converter types.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
Corniel: I added the converters that are defined. They are not used in this way yet. If you define a node as `Node<IOFile>` the default still would be `IOFile.Enpty`, and `Node<IOFile?>` would be `null.`

I'm think that MSBuild handles `null` and `string.Empty` the same, so in that case this is an improvement, but It's fair to say that this change takes more consideration end potentially should be postponed.

JohnVerheij: MSBuild's null/empty equivalence makes this effectively a no-op for typical usage. Happy with whichever way you decide.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
Corniel: Good suggestion.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
Corniel: I go for null.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
Corniel: Did that.

src/DotNetProjectFile.Analyzers/MsBuild/Conversion/TypeConverters.cs
Corniel: Added as well.

PR #801, opened 2026-05-14
Ignore <Target> nodes when reading for properties

### LINKED ISSUE #800: Proj0243 false positive: <GenerateSBOM> task element picked up instead of the property
When `<GenerateSBOM>true</GenerateSBOM>` is set in `Directory.Build.props` and a `<GenerateSBOM>` MSBuild task is invoked inside a `<Target>` in `Directory.Build.targets`, Proj0243 fires on the task element.

## Repro

`global.json`:

```json
{
  "sdk": {
    "version": "10.0.300",
    "rollForward": "latestFeature"
  }
}
```

`Directory.Packages.props`:

```xml
<Project>
  <PropertyGroup>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
  </PropertyGroup>
  <ItemGroup>
    <PackageVersion Include="Microsoft.Sbom.Targets" Version="4.1.5" />
  </ItemGroup>
  <ItemGroup>
    <GlobalPackageReference Include="DotNetProjectFile.Analyzers" Version="1.14.0" PrivateAssets="all" />
  </ItemGroup>
</Project>
```

`Directory.Build.props`:

```xml
<Project>
  <PropertyGroup>
    <Nullable>enable</Nullable>
    <GenerateSBOM>true</GenerateSBOM>
  </PropertyGroup>
</Project>
```

`Directory.Build.targets`:

```xml
<Project>
  <Target Name="GenerateSbomOnPublish" AfterTargets="Publish"
          Condition="'$(GenerateSBOM)' == 'true' AND '$(IsPackable)' != 'true'">
    <GenerateSBOM
      BuildDropPath="$(PublishDir)"
      PackageName="$(MSBuildProjectName)"
      PackageVersion="$(Version)"
      PackageSupplier="Example"
      NamespaceBaseUri="https://example.com/sbom" />
  </Target>
</Project>
```

`MyLib/MyLib.csproj`:

```xml
  <PropertyGroup>
    <IsPackable>true</IsPackable>
    <Version>1.0.0</Version>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Sbom.Targets" PrivateAssets="all" />
  </ItemGroup>
</Project>
```

Build with `dotnet build MyLib/MyLib.csproj` produces:

```
Directory.Build.targets(4,5): warning Proj0243: Enable SBOM generation with <GenerateSBOM> is 'true' or define the <IsPackable> node with value 'false'
```

`(4,5)` is the position of the `<GenerateSBOM` task element. Removing `Directory.Build.targets` from the repo silences the warning.

## Root cause

`MsBuildProject.Property<TNode>()` returns the first matching node from `SelfAndDirectoryProps()`, which yields `Directory.Build.targets` before `Directory.Build.props`. `<GenerateSBOM>` parses to the same `GenerateSBOM : Node<bool?>` type whether the parent is `<PropertyGroup>` or `<Target>`; the task element has no inner text so its `Value` is not `true`. `GenerateSbom.Register` then evaluates `property.Value is not true` on the task element and reports the diagnostic at its location.

Relevant files:

- `src/DotNetProjectFile.Analyzers/MsBuild/MsBuildProject.Property.cs`
- `src/DotNetProjectFile.Analyzers/MsBuild/MsBuildProject.cs`
- `src/DotNetProjectFile.Analyzers/MsBuild/GenerateSBOM.cs`
- `src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/GenerateSbom.cs`

Corniel:
Closes #800

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Generate_SBOM.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Rules/MS_Build/Generate_SBOM.cs
@@ -46,4 +46,30 @@ public class Guards
         </Project>
         """)
         .HasNoIssues();
+
+    [TestCase]
+    public void when_non_property_nodes_also_exist() => new GenerateSbom().ForInlineCsproj("""
+        <Project Sdk="Microsoft.NET.Sdk">
+
+          <PropertyGroup>
+            <TargetFramework>net10.0</TargetFramework>
+            <GenerateSBOM>true</GenerateSBOM>
+          </PropertyGroup>
+
+          <ItemGroup>
+            <PackageReference Include="Microsoft.Sbom.Targets" Version="4.1.5" />
+          </ItemGroup>
+
+          <Target Name="GenerateSbomOnPublish" AfterTargets="Publish">
+            <GenerateSBOM
+              BuildDropPath="$(PublishDir)"
+              PackageName="$(MSBuildProjectName)"
+              PackageVersion="$(Version)"
+              PackageSupplier="Example"
+              NamespaceBaseUri="https://example.com/sbom" />
+          </Target>
+
+        </Project>
+        """)
+        .HasNoIssues();
 }
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -83,7 +83,8 @@
       <![CDATA[
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
-
+v1.*
+- Ignore <Target> nodes when reading properties. (BUG)
 ]]>
     </ToBeReleased>
     <PackageReleaseNotes>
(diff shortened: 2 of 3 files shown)


JohnVerheij:
Verified locally against the #800 repro.

The fix is correct: `<Target>` contents are execution-time tasks, not
evaluation-time properties, so `Read<TNode>` shouldn't descend into them.
Skipping `Target` next to `ItemGroup` is the right minimal change and keeps
the existing denylist style.

Confirmed the spec is a genuine regression test. Applied to `main` without
the fix, `when_non_property_nodes_also_exist` fails (the reverse-order child
walk hits the `<Target>` before the `<PropertyGroup>`); with the fix it passes.

One nit: the `<remarks>` on `Read<TNode>` still says only "Skip ItemGroups as
they can not contain properties". Worth extending it so it mentions `Target`
too and the comment keeps matching the code.

Minor: the spec uses a single-file csproj, whereas #800 was split across
`Directory.Build.props` and `.targets`. The fix covers both since `Read<TNode>`
runs per file, so the single-file spec is fine; just noting the cross-file
form is the one originally reported.

Corniel:
@JohnVerheij If you agree with the changes, please approve the PR. :)

JohnVerheij (approved):
Doc remark updated as suggested; regression test covers the `<Target>` skip. LGTM.

PR #804, opened 2026-05-26
Configure cooldown for Dependabot updates

### LINKED ISSUE #803: Add Dependabot cooldown
Supply-chain compromises are on the rise so this is one thing we can do to give the larger community a chance to find issues both security wise and otherwise.

Docs: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

> The cooldown option is only available for version updates, not security updates. 

Can be different values per semver major/minor/patch at least for NuGet. (We should probably also pin GitHub Actions to specific commit sha (and enforce this at the repo level `Settings` -> `Actions` -> `General` -> Check `Require actions to be pinned to a full-length commit SHA`) and add Dependabot for version updates)

Suitable cooldown is likely 7-30 days any input appreciated here. Longer for major versions is probably good.

I can do a PR if this is something we want to do.

Frulfump:
Fix #803
Added cooldown periods for dependency updates in Dependabot configuration.
Also adds GitHub Actions Ecosystem with cooldown

Will follow up GitHub Actions SHA-pinning after this is merged (ideally it's also enforced as a repo policy afterwards)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,14 @@ updates:
     directory: "/" # Checks the entire repository
     schedule:
       interval: "weekly"
+    cooldown: # Defines a cooldown period for dependency updates, allowing updates to be delayed for a configurable number of days. The cooldown option is only available for version updates, not security updates. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-
+      semver-patch-days: 7
+      semver-minor-days: 21
+      semver-major-days: 31
+
+  - package-ecosystem: "github-actions" # Required
+    directory: "/" # Required (or directories). Location of package manifests. You must define the location of the package manifests for each package manager. For GitHub Actions, you do not need to set the directory to /.github/workflows. Configuring the key to / automatically instructs Dependabot to search the /.github/workflows directory, as well as the action.yml / action.yaml file from the root directory
+    schedule:
+      interval: "weekly"
+    cooldown: # Defines a cooldown period for dependency updates, allowing updates to be delayed for a configurable number of days. The cooldown option is only available for version updates, not security updates. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-
+      default-days: 31 # GitHub Actions doesn't support the SemVer cooldown (https://github.com/dependabot/dependabot-core/issues/14628) https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#configuration-of-cooldown
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,8 +9,8 @@ updates:
       semver-minor-days: 21
       semver-major-days: 31
 
-  - package-ecosystem: "github-actions" # Required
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,3 +4,14 @@ updates:
     directory: "/" # Checks the entire repository
     schedule:
       interval: "weekly"
+    cooldown: # Defines a cooldown period for dependency updates, allowing updates to be delayed for a configurable number of days. The cooldown option is only available for version updates, not security updates. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-
+      semver-patch-days: 7
+      semver-minor-days: 21
+      semver-major-days: 31
+
+  - package-ecosystem: "github-actions" # Required
```
Frulfump: Remove unnecessary comment
```suggestion
  - package-ecosystem: "github-actions"
```

.github/dependabot.yml
```diff
@@ -4,3 +4,14 @@ updates:
...
+    directory: "/" # Required (or directories). Location of package manifests. You must define the location of the package manifests for each package manager. For GitHub Actions, you do not need to set the directory to /.github/workflows. Configuring the key to / automatically instructs Dependabot to search the /.github/workflows directory, as well as the action.yml / action.yaml file from the root directory
```
Frulfump: Remove unnecessary comment
```suggestion
    directory: "/"
```

.github/dependabot.yml
```diff
@@ -4,3 +4,14 @@ updates:
     directory: "/" # Checks the entire repository
     schedule:
       interval: "weekly"
+    cooldown: # Defines a cooldown period for dependency updates, allowing updates to be delayed for a configurable number of days. The cooldown option is only available for version updates, not security updates. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-
+      semver-patch-days: 7
+      semver-minor-days: 21
```
Frulfump: Maybe a bit long?

CptWesley: Perhaps. I would also put this on 7 days personally, not sure how @Corniel feels about it.

Corniel: I think I agree with @CptWesley .

Corniel:
```suggestion
      semver-minor-days: 7
```

.github/dependabot.yml
```diff
@@ -4,3 +4,14 @@ updates:
     directory: "/" # Checks the entire repository
     schedule:
       interval: "weekly"
+    cooldown: # Defines a cooldown period for dependency updates, allowing updates to be delayed for a configurable number of days. The cooldown option is only available for version updates, not security updates. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-
+      semver-patch-days: 7
+      semver-minor-days: 7
+      semver-major-days: 31
+
+  - package-ecosystem: "github-actions"
```
CptWesley, line 12: I'm not entirely sure why we would want to introduce this package-ecosystem.

Frulfump: It would enable us to pin the version used by commit SHA and have dependabot update them. Currently we are more susceptible to supply chain attacks and bugs as we alway resolve latest (for a specific major) on the fly.

CptWesley (approved):

PR #814, opened 2026-06-01
<Content> to improve the SonarQube experience should not be packed

### LINKED ISSUE #813: <Content> to improve the SonarQube experience should not be packed
With #691 MSBuild files are now are included as `<Content>` to improve the SonarQube experience. As `<Content>` automatically included, MSBuild files are now packed when packages are created. These files should be added with `Pack="false"`.

Corniel:
Closes #813

### DIFF
--- a/src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
+++ b/src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
@@ -15,14 +15,17 @@
       Ensures that <Content> only included to satisfy SonarQube is not copied.
     
     * Visible="false"
-      Prevents from <Content> being shown it he IDE that was included by the user.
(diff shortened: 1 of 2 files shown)


Silvex78 (approved):

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
```diff
@@ -15,14 +15,17 @@
       Ensures that <Content> only included to satisfy SonarQube is not copied.
     
     * Visible="false"
-      Prevents from <Content> being shown it he IDE that was included by the user.
+      Prevents <Content> from being shown it he IDE that was included by the user.
```
Sjaaky:
```suggestion
      Prevents <Content> from being shown in the IDE that was included by the user.
```

Sjaaky (approved):
lgtm

PR #815, opened 2026-06-03
Centralize .NET SDK version definition

### LINKED ISSUE #811: CI is using outdated .NET SDK version
(it's also not strictly supported anymore and contains known vulnerabilities)
https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/blob/main/.github/workflows/build-publish.yml#L30C11-L30C37
Showing its using 10.0.101 

We should either use the latest 1xx feature brand or use the latest 3xx feature band and then we will have to switch to 4xx when 3xx goes out of support. 1xx is supported for the entirety of the .NET 10 life cycle. New feature bands can introduce breaking changes, new warnings etc. but it's usually no big deal to absorb them.

We should consider a `global.json` https://learn.microsoft.com/en-us/dotnet/core/tools/global-json
We could consider pinning the version used so we disable `rollForward` or we accept another `rollForward` policy like `latestFeature` and accept that there could be an inconsistency between CI and the version used locally by developers.
Dependabot can be used to handle version updates in `global.json` if we choose to pin it (maybe otherwise as well)

I can do a PR after we decide the fix

pmdevers:
Moves the .NET SDK version specification from the GitHub Actions workflow to a `global.json` file. This aligns with standard .NET project practices for managing SDK versions, ensures consistency across all projects in the repository, and simplifies future updates for both build and publish jobs.

fixes #811

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -26,8 +26,6 @@ jobs:
 
       - name: Setup dotnet
         uses: actions/setup-dotnet@v5.2.0
-        with:
-          dotnet-version: '10.0.101'
 
       - name: Build
         run: dotnet build dotnet-project-file-analyzers.slnx -c Release
@@ -61,6 +59,9 @@ jobs:
       contents: write
     needs: build
     steps:
+      - name: Setup dotnet
+        uses: actions/setup-dotnet@v5.2.0
+
       - name: Download Artifact
         uses: actions/download-artifact@v4
         with:
--- a/global.json
+++ b/global.json
@@ -1,0 +1,6 @@
+{
+  "sdk": {
+    "version": "10.0.101",
+    "rollForward": "latestPatch"
+  }
+}
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "10.0.101",
+    "version": "10.0.300",
     "rollForward": "latestPatch"
   }
 }


global.json
```diff
@@ -0,0 +1,6 @@
+{
+  "sdk": {
+    "version": "10.0.101",
```
Corniel:
```suggestion
    "version": "10.0.300",
```

While at it, lets bump the version too. See: [dotnet.microsoft.com](https://dotnet.microsoft.com/download/dotnet/10.0).

Corniel (approved):
LGTM!

.github/workflows/build-publish.yml
```diff
@@ -61,6 +59,9 @@ jobs:
       contents: write
     needs: build
     steps:
+      - name: Setup dotnet
```
Frulfump, line 62: Is there anything in the publish step that need the .NET SDK? And since this already has the `needs: build` isn't it then using the setup-dotnet from that action?

global.json
Frulfump: Agree, and especially since 10.0.101 contains known vulnerabilities it's unsuitable, if it was kept at 1xx it should have been the latest available but I think Corniels suggestion of 10.0.300 is best for us.

global.json
```diff
@@ -0,0 +1,6 @@
...
+    "rollForward": "latestPatch"
```
Frulfump, line 4: I need to think about the specific policy a bit, we didn't reach a consensus in the issue

Frulfump: I think `latestPatch` is fine unless we use lockfiles then the docs for global.json says this

>† When you use [package lock files](https://learn.microsoft.com/en-us/dotnet/core/install/upgrade#package-lock-files), set rollForward to disable so the SDK version and dependency graph stay in lockstep. For related issues, see https://github.com/dotnet/aspnetcore/issues/65061 and https://github.com/dotnet/sdk/issues/48795.

I think it will mean that setup dotnet will install the latest patch if we use the latest version of the action https://github.com/actions/setup-dotnet/releases/tag/v5.3.0

When 10.0.4xx releases we should update the version, Not sure if we should let Dependabot handle updates as well but we don't have to do that in this PR (if we choose to pin the version it's more appropriate to handle it here but not a strict requirement)

.github/workflows/build-publish.yml
```diff
@@ -61,6 +59,9 @@ jobs:
...
+        uses: actions/setup-dotnet@v5.2.0
```
Frulfump, line 63: That isn't the latest version https://github.com/actions/setup-dotnet/releases/tag/v5.3.0 and I think we want the support for latest* in that version

304NotModified (approved):
LGTM also

The setup-dotnet@ version could be in another PR (scope creep?)

Frulfump:
> LGTM also
> 
> The setup-dotnet@ version could be in another PR (scope creep?)

Yeah maybe, was just thinking about the claimed support in the latest release
> Support global.json's rollForward latest* variants

So that kind of makes the change in-scope but maybe not a strict requirement. It's patch Tuesday next Tuesday so I guess we will notice then if it installs the wrong version or not.

global.json
Corniel: I think we do too, but as @304NotModified suggested: let's do that in a follow-up PR.

.github/workflows/build-publish.yml
Corniel: I'm not sure. May be you can a create an issue to address that concern?

Frulfump: Sure, since it's being introduced in this PR I thought it would be the best place to adress before merging the PR. Created #817

PR #823, opened 2026-06-06
Proj0052: Prefer attributes over elements

### LINKED ISSUE #485: New Rule: Prefer Version attribute over element
Some (old converted) project files might contain a `<Version>` element instead of an attribute.

## Non-compliant
``` xml
<PackageReference Include="Qowaiv">
  <Version>8.4.3</Version>
</PackageReference>
```

## Compliant
``` xml
<PackageReference Include="Qowaiv" Version="8.4.3" />
```

Corniel:
The use of XML attributes is preferred over child XML elements for item metadata within MSBuild files. Attributes are more concise and easier to scan than nested elements, especially when configuring only a few metadata properties. This style also aligns with modern .NET SDK project conventions.

## Exceptions
* Structural Containers `<PropertyGroup>` and `<ItemDefinitionGroup>`  are excluded from this rule, as they are designed by MSBuild to use child  elements rather than attributes.
* Content item metadata containing multi-line text is automatically recognized and will not be flagged, as XML attributes are unable to properly preserve  line breaks.

Closes #485

### DIFF
--- a/src/DotNetProjectFile.Analyzers/Rule.cs
+++ b/src/DotNetProjectFile.Analyzers/Rule.cs
@@ -502,7 +502,7 @@ public static partial class Rule
         title: "Prefer attributes over elements",
         message: "Consider <{0}> as an attribute",
         description: "Attributes are more concise and readable than elements for (simple) values.",
-        tags: ["Clearity", "Readabillity"],
+        tags: ["Clearity", "Readability"],
         category: Category.Formatting);
 
     public static DiagnosticDescriptor DefineIsPackable => New(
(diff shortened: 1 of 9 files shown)


src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -497,6 +497,14 @@ public static partial class Rule
         tags: ["dependencies", "dependency", "file system", "UNIX", "case-consistent"],
         category: Category.Bug);
 
+    public static DiagnosticDescriptor PreferAttributesOverElements => New(
+        id: 0052,
+        title: "Prefer attributes over elements",
+        message: "Consider <{0}> as an attribute",
+        description: "Attributes are more concise and readable than elements for (simple) values.",
+        tags: ["Clearity", "Readabillity"],
```
LauraXiulan:
```suggestion
        tags: ["Clearity", "Readability"],
```

LauraXiulan (approved):
See comment

PR #833, opened 2026-06-22
Pack .targets and .props when explicitly configures in <None>

### LINKED ISSUE #828: build/*.targets no longer packed since 1.15.0 (Content glob forces Pack="false")
### Summary

Since 1.15.0, a project that references `DotNetProjectFile.Analyzers` and ships its own
`build/<id>.targets` as part of its NuGet package no longer includes that file in the
produced `.nupkg`. The targets file is excluded silently, with no warning. 1.14.1 packs
it correctly.

This breaks the standard mechanism where a package auto-imports MSBuild logic into
consumers via `build/<PackageId>.targets`.

### Cause

`build/DotNetProjectFile.Analyzers.targets` adds project files as `<Content … Pack="false">`
so they can be reported to SonarQube. 1.15.0 extended the glob set to include `**/*.targets`:

```xml
<ItemGroup Condition="'$(SonarQubeIntegration)' == 'true'">
  ...
  <Content Exclude="@(Content);**/bin/**/*;**/obj/**/*"
           CopyToOutputDirectory="never" Visible="false" Pack="false"
           Link="$([System.Guid]::NewGuid())" Include="**/*.targets" />
</ItemGroup>
```

A package's own `build/<id>.targets` sits under the project directory, so it matches this
glob and is marked `Pack="false"`. That `Content` item overrides the author's explicit
`<None Include="build/<id>.targets" Pack="true" PackagePath="build/" />`, so the file is
left out of the package. Because the glob is in an imported `.targets` (evaluated after the
project body), it cannot be overridden from the consuming project file.

### Minimal repro

`Repro.csproj`:

```xml
  <PropertyGroup>
    <IsPackable>true</IsPackable>
    <PackageId>Repro</PackageId>
    <Version>1.0.0</Version>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="DotNetProjectFile.Analyzers" Version="1.15.0" PrivateAssets="all" />
  </ItemGroup>
  <ItemGroup>
    <None Include="build/Repro.targets" Pack="true" PackagePath="build/" />
  </ItemGroup>
</Project>
```

`build/Repro.targets` (contents do not matter):

```xml
<Project />
```

Then:

```
dotnet pack -c Release
unzip -l bin/Release/Repro.1.0.0.nupkg
```

### Expected

`build/Repro.targets` is present in the package (as it is on 1.14.1).

### Actual

The package contains only `lib/`. `build/Repro.targets` is missing, and no warning is
emitted. Pinning `DotNetProjectFile.Analyzers` to 1.14.1 restores it.

### Workaround

Set `<SonarQubeIntegration>false</SonarQubeIntegration>`, which disables the content
ItemGroup above.

### Suggested fix

Exclude `build/**`, `buildTransitive/**`, and `buildMultiTargeting/**` from the SonarQube
content globs (those are package-author assets, not analysis inputs), or skip files the
project already includes with `Pack="true"`. A `.targets`/`.props` shipped under `build/`
is part of the package, so forcing `Pack="false"` on it changes the produced package.

### Affected versions

Regressed in 1.15.0. 1.14.1 is unaffected.

### Introduced by

The `**/*.targets` glob entry was added in #822 ("Hide folders only containing invisible
`<Content>`"), first shipped in 1.15.0. 1.14.1 globs `**/*.props` but not `**/*.targets`,
so packages shipping `build/<id>.props` were already affected before 1.15.0; 1.15.0 extends
the same issue to `build/<id>.targets`.

Corniel:
Closes #828


src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
```diff
@@ -18,20 +18,23 @@
...
-    <Content Exclude="@(Content);**/bin/**/*;**/obj/**/*" CopyToOutputDirectory="never" Visible="false" Pack="false" Link="$([System.Guid]::NewGuid())" Include="**/*.slnx" />
+    <None
+       Include="**/*.??proj;**/*.config;**/*.props;**/*.targets;**/*.resx;**/*.slnx"
+       Exclude="@(None);@(Content);**/bin/**/*;**/obj/**/*" />
+    <Content
+       Include="@(None)"
+       CopyToOutputDirectory="never"
+       Visible="false"
+	     Pack="%(None.Pack)"
+       Link="$([System.String]::Copy('%(Identity)').Replace('\', '_').Replace('\', '_'))" />
```
JohnVerheij: Why is this not the same as stated on line 24? I think the second `Replace` should be `.Replace('/', '_')`

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
```diff
@@ -18,20 +18,23 @@
...
-    <Content Exclude="@(Content);**/bin/**/*;**/obj/**/*" CopyToOutputDirectory="never" Visible="false" Pack="false" Link="$([System.Guid]::NewGuid())" Include="**/*.resx" />
-    <Content Exclude="@(Content);**/bin/**/*;**/obj/**/*" CopyToOutputDirectory="never" Visible="false" Pack="false" Link="$([System.Guid]::NewGuid())" Include="**/*.slnx" />
+    <None
+       Include="**/*.??proj;**/*.config;**/*.props;**/*.targets;**/*.resx;**/*.slnx"
+       Exclude="@(None);@(Content);**/bin/**/*;**/obj/**/*" />
+    <Content
+       Include="@(None)"
+       CopyToOutputDirectory="never"
+       Visible="false"
+	     Pack="%(None.Pack)"
```
JohnVerheij: This pulls in the consumer's own None items too, so with `Pack="%(None.Pack)"` their other project files (.csproj/.config/.props/.targets/.resx/.slnx) get an empty Pack and land in `content/` and `contentFiles/` (IncludeContentInPack defaults to true), and `build/MyPkg.targets` gets packed twice (NU5118). Setting it back to `Pack="false"` just brings back #828 (NU5017).

I tested with a small consumer (class lib, IncludeBuildOutput=false, one `<None Include="build/MyPkg.targets" Pack="true" PackagePath="build/">`). What packed cleanly was globbing into a private item and forcing Pack=false on only those, leaving the consumer's items alone:

```xml
<ItemGroup Condition="'$(SonarQubeIntegration)' == 'true'">
  <_SonarQubeContent
     Include="**/*.??proj;**/*.config;**/*.props;**/*.targets;**/*.resx;**/*.slnx"
     Exclude="@(None);@(Content);**/bin/**/*;**/obj/**/*" />
  <Content
     Include="@(_SonarQubeContent)"
     CopyToOutputDirectory="never" Visible="false" Pack="false"
     Link="$([System.String]::Copy('%(Identity)').Replace('\', '_').Replace('/', '_'))" />
</ItemGroup>
```

Then the package only has `build/MyPkg.targets`.

JohnVerheij (commented):
Thanks for picking this up. Two notes inline.

The main one: as written this fixes the dropped targets but leaks the consumer's own project files into `content/` and `contentFiles/`, so it shouldn't merge as-is.  The Link one is minor.

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
Corniel: That is a mistake. I fixed it.

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
Corniel: I thought about a construction like this. By doing so, however, Files that have been added as `<None>` explicitly, will not be reported by Sonar. I've now updated to `Pack="$([System.String]::Equals('%(None.Pack)', 'true', System.StringComparison.OrdinalIgnoreCase) ? 'true' : 'false')"` which seems to work locally.

JohnVerheij: One option is AdditionalFiles instead of Content. It has no pack semantics, so a file can be both a packed `<None>` and an analyzed `AdditionalFile` with no collision. In a repro that packs clean (only the explicitly-packed targets, no leak, no NU5017/NU5118) and the files still reach the analyzer.

Haven't checked how Sonar indexes AdditionalFiles vs Content, so that side you'd know better than me.

Corniel: Sonar does not check AdditionalFiles. So that will not solve the issue.

JohnVerheij: Another approach that keeps SonarQubeIntegration on and packs clean: keep the Content for Sonar but remove it from the pack items before pack runs.

```xml
<Content Include="@(None)" Visible="false" CopyToOutputDirectory="never" SonarQubeContent="true" Link="..." />

<Target Name="_StripSonarContent" BeforeTargets="_GetPackageFiles">
  <ItemGroup>
    <Content Remove="@(Content->WithMetadataValue('SonarQubeContent','true'))" />
  </ItemGroup>
</Target>
```

Verified against a consumer that packs `build/x.targets` via a `<None>` with `Pack="true"` and `PackagePath="build/"`, IncludeBuildOutput=true: the package ends up with just `lib/` and `build/x.targets`, no project files in `content/`/`contentFiles/`, no NU5017/NU5118.

The strip only runs on pack (BeforeTargets="_GetPackageFiles"), so a Sonar build still sees the Content unchanged. I haven't run it through SonarCloud myself, but since the strip never runs during a Sonar build, indexing should be the same as today. Worth a quick check in your pipeline.

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -88,6 +88,8 @@ ToBeDecided:
     </ToBeReleased>
     <PackageReleaseNotes>
       <![CDATA[
+v1.15.1
+- Pack .targets and .props when explicitly configures in <None>. (BUG)
```
CptWesley: not gramatically correct

CptWesley (approved):

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
Corniel: I'd like this suggestion. I've made a small change in the `<Target>`. Only execute it if both `Packing` and `SonarQubeContent` are enabled, and use `BeforeTargets="GenerateNuspec"`.

JohnVerheij: The trigger needs to be _GetPackageFiles, not GenerateNuspec.

_GetPackageFiles is the target that gathers the items into the package, and GenerateNuspec depends on it. So a strip hooked to GenerateNuspec runs after the gather already happened, and removing the Content from @(Content) at that point doesn't pull it back out of what was already collected. A consumer pack still gets:

```
content/MyPkg.csproj
contentFiles/any/net10.0/MyPkg.csproj
content/build/MyPkg.targets
+ NU5118 on build/MyPkg.targets
```

BeforeTargets="_GetPackageFiles" runs the strip before the gather, so the items are gone before they're collected and the package is clean (just lib/ and build/MyPkg.targets).

The own-package CI doesn't catch it because the leak only shows up in a consumer that ships its own build files.

Corniel: Done

JohnVerheij: Still leaking. The trigger is right now, but the `Condition="'$(IsPacking)' == 'true'"` on the target never holds during pack. `$(IsPacking)` is empty; the SDK sets `$(_IsPacking)`. A probe at `_GetPackageFiles` prints:

```
IsPacking=[] _IsPacking=[true]
```

So `_StripSonarContent` is skipped and a consumer pack still gives:

```
content/MyPkg.csproj
content/build/MyPkg.targets
contentFiles/any/net10.0/MyPkg.csproj
contentFiles/any/net10.0/build/MyPkg.targets
+ NU5118 on build/MyPkg.targets
```

Dropping the condition fixes it. The target already runs only during pack via `BeforeTargets="_GetPackageFiles"`, so the guard adds nothing. The package is then just `lib/` and `build/MyPkg.targets`.

Every version of this fix has built green here but still left a consumer pack broken, because nothing in CI packs a consumer that ships `build/*.targets`. The package test only inspects this repo's own package. A one-project fixture that packs a consumer and asserts no `content/` or `contentFiles/` entries and no NU5118 would catch all of these. Can share the repro if useful.

Corniel: I'm thinking of using Buildalizer to add a test that checks this behavior automatically. Thanks for the reports.

Corniel:
@JohnVerheij I've been able to add tests that check the actual behavior. Some Content is now available twice, so I have to dive into why that is.

JohnVerheij:
`577f781` (Simplified) removed the `_StripSonarContent` target and `d571496` did not put it back, so the SonarQube `Content` items now reach pack. A consumer that ships a build file through a packable `None` then leaks its own files into the package.

Repro: a class library with default `SonarQubeIntegration`, a packable `None` for `build/MyPkg.targets`, referencing this build of the analyzer. The produced package contains:

```
build/MyPkg.targets
content/MyPkg.csproj
content/build/MyPkg.targets
contentFiles/any/net10.0/MyPkg.csproj
contentFiles/any/net10.0/build/MyPkg.targets
lib/net10.0/MyPkg.dll
```

plus NU5118 because `build/MyPkg.targets` is added twice. The exact extra files depend on what the consumer has, but the project file and the build file (twice) are always there.

Restoring the target fixes it:

```xml
<Target Name="_StripSonarContent" BeforeTargets="_GetPackageFiles">
  <ItemGroup Condition="'$(SonarQubeIntegration)' == 'true'">
    <Content Remove="@(Content->WithMetadataValue('SonarQubeContent','true'))" />
  </ItemGroup>
</Target>
```

With it the package is `build/MyPkg.targets` + `lib/net10.0/MyPkg.dll` only, no leak and no NU5118. `BeforeTargets="_GetPackageFiles"` keeps it pack-only, so the `Content` is still present during build for Sonar and gone by `GenerateNuspec` (9 items during build, 0 at pack in the repro).

`Pack="false"` on the `Content` is not an alternative: it overrides the consumer's own `None` `Pack="true"` for the same file, so the build file drops out of the package. With other content present it ships silently without the build file; when the build file is the only asset it fails with NU5017.

The comment above the ItemGroup still lists `Pack = false` as the mechanism, but there is no `Pack` attribute on the item.

Corniel:
@ JohnVerheij Can you look again?

JohnVerheij:
Still leaks at `d7395695` (the consumer pack has the `content/` and `contentFiles/` copies plus NU5118). `_StripSonarContent` is not in the current head.

I have suggested that target a few times now: it was in `effa458`, and again in my comment above. Is there a reason you do not want the strip approach? If you would prefer another mechanism I can look at it. Otherwise re-adding the `effa458` target makes the consumer pack clean.

CI looks green because the `Package` test that reads the nupkg is `[Explicit]` outside `DEBUG`, so the Release run skips it.

Corniel:
@JohnVerheij I'm not against the strip approach, not at all. I added a test that (locally) suggests that this is sufficient. The reason to put on `[Explicit]` is that the package is not available on disk at the place where I think it was.

I'd like to have a(n automated) test that proofs that it work as it should, so that it can not be broken unintentionally, in the future.

JohnVerheij:
Packing `CompliantCSharpPackage` at the current head (`d7395695`) still leaks the package's own sources and project files. `Package()` passes anyway: it only asserts four files are present, and it is `[Explicit]`, so CI does not run it.

On top of the expected payload, the produced package contains 14 extra entries:

```
content/CompliantCSharpPackage.csproj
content/Messages.resx
content/README.md
content/build/CompliantCSharpPackage.props
content/build/CompliantCSharpPackage.targets
content/compliant-package.slnx
content/packages.lock.json
contentFiles/any/net10.0/CompliantCSharpPackage.csproj
contentFiles/any/net10.0/Messages.resx
contentFiles/any/net10.0/README.md
contentFiles/any/net10.0/build/CompliantCSharpPackage.props
contentFiles/any/net10.0/build/CompliantCSharpPackage.targets
contentFiles/any/net10.0/compliant-package.slnx
contentFiles/any/net10.0/packages.lock.json
```

A `NotContain("content/")` style guard does not work here: the project ships `logo_128x128.png` as a real `Content` item, so a correct package still has `content/logo_128x128.png` and `contentFiles/any/net10.0/logo_128x128.png`. Asserting the absence of a prefix would fail on a correct package.

The reliable check is to pin the full payload. One assertion then catches the leak, a missing build asset (the original #828 symptom), and a double-added file:

```csharp
[Test]
public void Package()
{
    using var ctx = BuildalyzerContext.ForProject("CompliantCSharpPackage/CompliantCSharpPackage.csproj");

    var options = new EnvironmentOptions() { DesignTime = false };
    options.Arguments.Add("-p:GeneratePackageOnBuild=true");

    ctx.Analyzer.Build(options);

    var package = Path.Combine(ctx.Location.Directory!.FullName, "bin/Debug/CompliantCSharpPackage.1.0.0.nupkg");

    // NuGet and SBOM scaffolding filtered out so the assertion pins the actual payload.
    var payload = Nupkg.Read(new(package))
        .Where(e => !e.StartsWith("_rels/", StringComparison.Ordinal)
            && !e.StartsWith("_manifest/", StringComparison.Ordinal)
            && e != "[Content_Types].xml"
            && !e.EndsWith(".nuspec", StringComparison.Ordinal));

    payload.Should().BeEquivalentTo([
        "README.md",
        "logo_128x128.png",
        "build/CompliantCSharpPackage.props",
        "build/CompliantCSharpPackage.targets",
        "content/logo_128x128.png",
        "contentFiles/any/net10.0/logo_128x128.png",
        "lib/net10.0/CompliantCSharpPackage.dll",
        "lib/net10.0/CompliantCSharpPackage.xml",
    ]);
}
```

The expected list is what the package contains once the SonarQube content no longer reaches pack. I captured it by packing the same project with `SonarQubeIntegration=false`. At the current head the assertion fails with the 14 extra entries above; with the strip in place it passes.

Honest tradeoff: pinning the exact payload is brittle to intentional packaging changes. Adding a packed file, another target framework, an SDK change to the `contentFiles` layout, or toggling the SBOM will fail it until the expected list is updated. That is the cost, and it is real. The upside is the same property in reverse: nothing changes the produced package without a test catching it. If the update friction matters, the same assertion as an approval test (write the sorted manifest to a verified file and accept on change) keeps the coverage with a one-step update.

Two more notes:

- Drop the `[Explicit]` so the check runs in CI. The build is Debug through Buildalyzer regardless of the outer configuration, so the `bin/Debug` path stays valid.
- The logo appears three times in the payload (root, `content/`, `contentFiles/`) because it is both a packed `None` and a `Content` item. That is existing behavior, unrelated to the strip, and part of the expected set above.

JohnVerheij:
Your debug enumerate will show it: in CI the package lands under `bin/Release`, not `bin/Debug`. That is the cause. The test reads `bin/Debug/CompliantCSharpPackage.1.0.0.nupkg`, so `Nupkg.Read` throws `FileNotFoundException` before it reaches the assertion. It is the `[Explicit]` reason you mentioned earlier: the package is on disk, just not under `bin/Debug`.

The strip target and the payload assertion are both fine. I packed `CompliantCSharpPackage` on Linux and the filtered payload matches your `BeEquivalentTo` list exactly (0 missing, 0 extra).

Locating the package instead of hardcoding the configuration folder fixes it:

```csharp
var package = ctx.Location.Directory!
    .EnumerateFiles("CompliantCSharpPackage.1.0.0.nupkg", SearchOption.AllDirectories).Single();
var payload = Nupkg.Read(package).Where(e => !Ignore(e));
```

`BuildalyzerContext` deletes `bin`/`obj` before the build, so there is exactly one `.nupkg` to find, whether it lands in `bin/Debug` or `bin/Release`. I ran the Package test on Linux with this change and it passes (`-c Release`, `--no-build`). (`"*.nupkg"` instead of the full name works too if you want it version-agnostic.)

Corniel:
@JohnVerheij Unfortunately, it does not:
>  System.IO.DirectoryNotFoundException : Could not find a part of the path '/home/runner/work/dotnet-project-file-analyzers/dotnet-project-file-analyzers/projects/CompliantCSharpPackage/bin/Release/CompliantCSharpPackage.1.0.0.nupkg'.

JohnVerheij:
The `artifacts` output path is the new problem. With `PackageOutputPath=artifacts` the pack output, including the `*.temp` staging folder, lands inside the project, so the SonarQube `Content` glob (`@(None);@(AdditionalFiles)`) sweeps it all up. That is the 13 extra `SonarQubeContent` items in `With_defaults` (the `artifacts/...temp/...` entries). (The new line also trips `Proj1700`, an unrelated indentation nit; it disappears when you drop `PackageOutputPath`.) `bin`/`obj` are excluded from that glob, so a package left in `bin` does not pollute the content set.

The path that avoids both is to drop `PackageOutputPath` and locate the nupkg under `bin` instead of assuming the folder:

```csharp
var package = ctx.Location.Directory!
    .EnumerateFiles("CompliantCSharpPackage.1.0.0.nupkg", SearchOption.AllDirectories).Single();
var payload = Nupkg.Read(package).Where(e => !Ignore(e));
```

This passed for me locally (`-c Release`). I still cannot reproduce the CI `FileNotFound` on my side, so it is worth a CI run to confirm, but at least it keeps the output out of the glob.

PR #838, opened 2026-07-01
Improve file cache

CptWesley:
Current cache implementation:
- Potentially parses file multiple times in parallel. (Solved in this PR)
- `LastWriteTimeUtc` is quite coarse and might miss multiple edits that come in repeatedly. (Solved in this PR)
- If file exists, but `LastWriteTimeUtc` comes back as `null`, the file is treated as unedited if this also happened the previous time. (Solved in this PR)
- Cache might potentially cache entire or parts of compilation due to internal/private references in `SourceText`/`AdditionalText`.
- If user switches to different solution inside same IDE instance, the cached values might linger forever. (Not solved)

The last two remaining issues can be mitigated/impact reduced by making the cache evict old/unused entries. But I wasn't sure if it was worth the time/effort investment.

### DIFF
--- a/specs/DotNetProjectFile.Analyzers.Specs/Caching/FileCache_specs.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/Caching/FileCache_specs.cs
@@ -57,6 +57,24 @@ public class TryGetOrUpdate
     }
 
     [Test]
+    public async Task Updates_existing_entry_when_length_changes_but_write_time_is_equal()
+    {
+        using var file = await TempFile.CreateText("Hello, World!");
+        var stamp = File.GetLastWriteTimeUtc(file.Path);
+
+        var cache = new FileCache<Entry>();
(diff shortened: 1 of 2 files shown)


Corniel (approved):
LGTM! How did you spot this issue if I may ask?

CptWesley:
I was running into a memory leak issue in a project of mine, where DevHub.exe was eating up all my RAM over time, requiring me to restart my IDE every hour or so. My suspicions were on a source code generator of mine (because I do complex stuff) and this project (because it does not-recommended-by-Microsoft things like IO).

I couldn't necessarily find issues in either myself, but then resorted to using Claude Code to try and find potential issues in my source code generator, which it did. Solving those fixed my big memory leak issue, but I still noticed DevHub.exe's RAM usage growing over time.

So, I also asked Claude Code to look into this project and it found these issues. Where the main issue it found (which I attempt to solve with this PR) was the file cache returning stale versions of the parsed files in certain scenarios.

PR #841, opened 2026-07-07
Add <EmbeddedResource> to <AdditionalFiles> once

### LINKED ISSUE #840: Duplicate 'Content' items were included error
When a project file already has a resx file included as `<AdditionalFiles>` the following error occurs:

##[error]/opt/hostedtoolcache/dotnet/sdk/10.0.301/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.DefaultItems.Shared.targets(213,5): Error NETSDK1022: Duplicate 'Content' items were included. The .NET SDK includes 'Content' items from your project directory by default. You can either remove these items from your project file, or set the 'EnableDefaultContentItems' property to 'false' if you want to explicitly include them in your project file. For more information, see https://aka.ms/sdkimplicititems. The duplicate items were: '........resx'

Corniel:
When adding an embedded resource as additional file yourself, an error occurred.

Closes #840

### DIFF
--- a/projects/CompliantCSharpPackage/CompliantCSharpPackage.csproj
+++ b/projects/CompliantCSharpPackage/CompliantCSharpPackage.csproj
@@ -80,6 +80,8 @@ v1.0.0
   </ItemGroup>
 
   <ItemGroup>
+    <!-- This is not needed but should not add it twice -->
+    <AdditionalFiles Include="*.resx" />
     <AdditionalFiles Include="*.slnx" />
   </ItemGroup>
   
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -79,18 +79,18 @@
   </ItemGroup>
 
   <PropertyGroup Label="Version and release notes">
-    <Version>1.15.1</Version>
+    <Version>1.15.2</Version>
     <ToBeReleased>
       <![CDATA[
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
-v1.*
-- Add <EmbeddedResource> to <AdditionalFiles> once. (BUG)
-- Improve file cache. 
 ]]>
     </ToBeReleased>
     <PackageReleaseNotes>
       <![CDATA[
+v1.15.2
+- Add <EmbeddedResource> to <AdditionalFiles> once. (BUG)
+- Improve file cache. 
 v1.15.1
 - Pack .targets and .props when explicitly configured in <None>. (BUG)
 v1.15.0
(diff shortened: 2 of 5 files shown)


JohnVerheij (commented):
I could reproduce #840. It needs two ingredients: a Web SDK project (`EnableDefaultContentItems` is only `true` there, which is what arms the check at Microsoft.NET.Sdk.DefaultItems.Shared.targets line 213) and an explicit `<AdditionalFiles Include="Foo.resx" />` in the project file body.

The chain is:

1. `AdditionalFiles.props` adds `**/*.resx` to `<AdditionalFiles>`, and the user's project file adds the same file again, so it is in `@(AdditionalFiles)` twice.
2. `DotNetProjectFile.Analyzers.targets` projects `@(None);@(AdditionalFiles)` into `<Content>` for the SonarQube integration (`SonarQubeIntegration` defaults to `true`).
3. The file ends up in `<Content>` twice and the Web SDK fails with NETSDK1022.

I tested this patch against that repro (by overwriting the props in a local package cache) and unfortunately the error still occurs, for two independent reasons:

1. `Exclude="$(AdditionalFiles)"` is a property reference. No such property exists, so it expands to empty and excludes nothing. An item reference would be `@(AdditionalFiles)`. The pre-existing Directory.Build.props/Directory.Build.targets/Directory.Packages.props entries in this file have the same issue.
2. Even with `@(AdditionalFiles)` it would not help here (also verified): NuGet imports `build/*.props` before the project body, so at the moment the glob is evaluated the user's `<AdditionalFiles>` entry does not exist yet. An `Exclude` in the props can only see items added by earlier imports.

What did work in my repro is moving the glob to the `.targets`, which is imported after the project body and therefore sees the user's items:

```xml
<!-- AdditionalFiles.props: remove the **/*.resx glob -->

<!-- DotNetProjectFile.Analyzers.targets, before the SonarQube ItemGroup -->
<ItemGroup>
  <AdditionalFiles Include="**/*.resx" Exclude="@(AdditionalFiles)" />
</ItemGroup>
```

With that change the repro builds with 0 errors and `Foo.resx` occurs exactly once in `@(AdditionalFiles)` (the user's own entry, so their metadata is preserved).

It may also be worth deduplicating the SonarQube `<Content>` projection itself as a second line of defense, since a duplicate `<AdditionalFiles>` entry from any other source would trigger the same NETSDK1022 in web projects.

Minimal repro (SDK 10.0.301):

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
  </PropertyGroup>
  <ItemGroup>
    <AdditionalFiles Include="Foo.resx" />
  </ItemGroup>
  <ItemGroup>
    <PackageReference Include="DotNetProjectFile.Analyzers" Version="1.15.1" PrivateAssets="all" />
  </ItemGroup>
</Project>
```

Corniel:
@JohnVerheij Thanks for the analysis. I applied your suggestion. I also extended a test that proved that my first approach did not work, and yours does.

JohnVerheij (approved):
Verified against the Web SDK repro from #840: builds clean, the resx is in AdditionalFiles and Content exactly once. The Directory.Build.* excludes are fixed too now that they use @() and evaluate after the project body.

PR #843, opened 2026-07-07
Skip SonarQube <Content> when building inside Visual Studio

### LINKED ISSUE #842: Unexpected problem loading one or more project
With the latest version of the project file analyzers a lot of our solutions give an error while loading projects:

LimitedFunctionality
System.ArgumentException: FindExcludedItem only accepts folder as the first parameter.
Parameter name: folder
   at Microsoft.VisualStudio.ProjectSystem.Designers.PhysicalProjectTreeProvider.FindExcludedItem(IProjectTree folder, String leafName)
   at Microsoft.VisualStudio.ProjectSystem.Designers.PhysicalProjectTreeProvider.HandleAddedItems(IProjectTree tree, IImmutableDictionary`2 projectData, IProjectChangeDescription change, IEnumerable`1 addedItems, ConfiguredProjectExports configuredProjectExports)
   at Microsoft.VisualStudio.ProjectSystem.Designers.PhysicalProjectTreeProvider.OnProjectSubscriptionChanged(IProjectTree originalTree, ConfiguredProjectExports configuredProjectExports, ConfiguredProjectExports originConfiguredProjectExports, PhysicalProjectTreeInputData updateInput, CancellationToken cancellationToken)

JohnVerheij:
Fixes #842.

The SonarQube `<Content>` projection introduced in 1.15.1 crashes the Visual Studio project system. The flattened `Link` only replaces slashes, so items from outside the project directory (`Directory.Build.props`, `Directory.Build.targets`, `Directory.Packages.props`, included as absolute paths) keep the drive colon in their link: `C:_Users_..._Directory.Build.props`.

When such an item reaches VS as part of a project update (for example right after bumping the package version, which is how #842 was hit):

1. `PhysicalProjectTreeProvider.FindOrCreateItem` computes `Path.GetDirectoryName("C:_Users_...")`, which for a drive-relative string is `"C:"`.
2. `FindOrCreateFolder` refuses a first segment with a volume separator and returns null.
3. `HandleAddedItems` then leaves its folder cursor on the previously added file node, and the next item in the batch throws `ArgumentException: FindExcludedItem only accepts folder as the first parameter`.

That is the exact stack in #842. It only fires on incremental updates, never on a clean first load, and it depends on the order of the added-items batch. Small test projects sometimes survive, real solutions with many files practically never do. There is a second variant from the same item group: the `$(MSBuildProjectFile)` entry gets `Link="MyProject.csproj"`, which collides with the hidden project file node VS keeps in its tree and fails in `RenameDuplicateItemInSameFolder`.

Repro: a project with a `Directory.Build.props` above it and a few loose files, referencing 1.14.1. Open the solution in VS, then bump the reference to 1.15.1 with the solution open. Result is the "Visual Studio ran into an unexpected problem with one or more projects" bar with the #842 stack in the ActivityLog.

The fix skips the item group when `$(BuildingInsideVisualStudio)` is true. VS never runs the SonarQube scanner (that happens on command line builds), so the items serve no purpose there. Verified against the repro with a patched package on VS 2026 (18.0): the same 1.14.1 to 1.15.1 bump that crashes with the stock targets loads clean with the gated one, both exception variants gone, while `dotnet build` still gets all items. The new spec mirrors `Does_not_add_content_with_SonarQubeIntegration_disabled` and fails without the targets change.

### DIFF
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -84,6 +84,7 @@
       <![CDATA[
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
+- Skip SonarQube <Content> when building inside Visual Studio. (BUG)
 ]]>
     </ToBeReleased>
     <PackageReleaseNotes>
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -84,7 +84,7 @@
       <![CDATA[
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
-- Skip SonarQube <Content> when building inside Visual Studio. (BUG)
+- Skip SonarQube <Content> when building inside Visual Studio, Rider, or ReSharper. (BUG)
 ]]>
     </ToBeReleased>
     <PackageReleaseNotes>
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -84,12 +84,12 @@
       <![CDATA[
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
-- Skip SonarQube <Content> when building inside Visual Studio, Rider, or ReSharper. (BUG)
 ]]>
     </ToBeReleased>
     <PackageReleaseNotes>
       <![CDATA[
 v1.15.2
+- Skip SonarQube <Content> when building inside Visual Studio, Rider, or ReSharper. (BUG)
 - Add <EmbeddedResource> to <AdditionalFiles> once. (BUG)
 - Improve file cache. 
 v1.15.1
--- a/specs/DotNetProjectFile.Analyzers.Specs/MS_Build/Package_assets_specs.cs
+++ b/specs/DotNetProjectFile.Analyzers.Specs/MS_Build/Package_assets_specs.cs
@@ -242,6 +242,27 @@ public class Builds
     }
 
     [Test]
+    public void Does_not_add_content_when_building_inside_Visual_Studio()
+    {
+        using var ctx = BuildalyzerContext.ForProject("CompliantCSharpPackage/CompliantCSharpPackage.csproj");
+
+        var options = new EnvironmentOptions() { DesignTime = false };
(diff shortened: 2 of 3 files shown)


Corniel (approved):
LGTM! Please add a short description in the `<ToBeReleased>` notes too.

Corniel:
@JohnVerheij Did you consider to extend this to VSCode and/or Rider IDE?

JohnVerheij:
Yes, and it turned out to be needed. Rider does not crash like VS, but it is affected: with 1.15.1 the invisible linked `<Content>` duplicates suppress the real files in Rider's solution view (the files exist on disk and as `<None>`, but do not show). Measured with a probe target on the #842 repro:

* Rider sets `BuildingInsideVisualStudio=true` only intermittently (the first open of a solution here, not on later loads), so the VS condition alone does not reliably cover it.
* `BuildingByReSharper=true` is set in every Rider phase captured: project loading, design-time builds, and real builds. ReSharper uses the same property.
* With the condition extended by `'$(BuildingByReSharper)' != 'true'`, the hidden files reappear in Rider's solution view.

Pushed that: condition extended plus a spec mirroring the VS one.

VS Code is a different story: there is no IDE build to detect. Everything it runs is a plain `dotnet build`, which is exactly the shape of build the SonarQube scanner uses, so a VS Code condition would also filter real scanner builds.

Corniel:
Gemini suggested `$(TERM_PROGRAM) == 'vscode'` but I did not test that.

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
```diff
@@ -84,6 +84,7 @@
       <![CDATA[
 ToBeDecided:
 - Proj0037: Exclude runtime when all assets are private. (NEW RULE)
+- Skip SonarQube <Content> when building inside Visual Studio, Rider, or ReSharper. (BUG)
```
Corniel: Now that #841 has been merged. This should be added tot 1.15.2, a version I'd like to ship this week.

JohnVerheij: Done. Rebased on main and moved the note to the v1.15.2 section.

JohnVerheij:
`TERM_PROGRAM` marks the terminal a command was typed in, not what the build is for. VS Code sets it to `vscode` in its integrated terminal, and MSBuild picks environment variables up as properties (verified with a probe target: `dotnet build` with the variable set sees `$(TERM_PROGRAM)` as `vscode`). So the condition would trigger for everything typed in that terminal, including a local scanner run (`dotnet sonarscanner begin`, `dotnet build`, `end`). That is exactly the build the Content items exist for, so the gate would filter real scanner builds. The builds VS Code runs itself come from the extension host, not the integrated terminal, so those would not carry the variable. Wrong builds on both sides, and there is no observed VS Code symptom to fix, so I would leave VS Code out.

Corniel (approved):

PR #845, opened 2026-07-08
Reimplement Grammr

Corniel:
I never fully finished my `Grammr` idea. The main Idea now is to use a `ref struct SourceReader`  (similar to `Utf8JsonReader`) that allows different parsers to read sections and conditionally to update the reference (on a match). The `Lexer`'s, allow to describe matches, and can be used to add tokens to the `TokenStream` of the `SourceReader`, but this is not required.

The speed also has improved. The previous implementation (INI) was 4.7 ys/LoC for the 1220 file. So, the new one is more than 7 times faster.

## Parsing INI files
|     File | Mean       | Gen0     | Gen1     | Gen2     | Allocated  |      Speed |
|---------:|-----------:|---------:|---------:|---------:|-----------:|-----------:|
|   27 LoC |   7.611 us |   1.6861 |   0.0610 |        - |   20.70 KB | 282 ns/LoC |
|   36 LoC |   7.957 us |   1.5411 |   0.0458 |        - |   18.91 KB | 221 ns/LoC |
| 1220 LoC | 754.667 us | 272.4609 | 272.4609 | 272.4609 | 2168.57 KB | 619 ns/LoC |

## Parsing .gitignore files
|     File | Mean     | Gen0    | Gen1   | Allocated |      Speed |
|---------:|---------:|--------:|-------:|----------:|-----------:|
|  344 LoC | 112.0 us |  9.2773 | 2.1973 | 114.00 KB | 326 ns/LoC |
|  606 LoC | 222.7 us | 13.4277 | 4.3945 | 166.49 KB | 367 ns/LoC |


src/DotNetProjectFile.Analyzers/Ini/IniFileParser.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                AddSection(reader, tree);
+                iniHeader = h;
+            }
+            else
+            {
+                reader.Cons(ws);
+                reader.Cons(comment);
+                reader.Cons(eol);
+            }
```
JohnVerheij: I think this can loop forever. When a line matches neither `Entry` nor `Header`, none of `Cons(ws)` / `Cons(comment)` / `Cons(eol)` is guaranteed to consume anything (`ws`/`eol` return null on a non-match, `comment` is optional so it matches zero-width), so `while (!reader.EOS)` never advances.

Each of these hangs (well-formed input parses fine):

```text
"@\n"
"foo bar\n"
"= value\n"
"root\n"          // a typo'd `root = true`
"[No Header\n"    // the fixture line in Rules/INI/INI_syntax

File: src/DotNetProjectFile.Analyzers/Analyzers/Ini/IniSyntaxAnalyzer.cs [hunk 2]
```diff
@@ -9,7 +9,7 @@ public sealed class IniSyntaxAnalyzer() : IniFileAnalyzer(
     /// <inheritdoc />
     protected override void Register(IniFileAnalysisContext context)
     {
-        foreach (var diagnostic in context.File.Syntax.GetDiagnostics())
+        foreach (var diagnostic in context.File.GetDiagnostics())
```
JohnVerheij, line 12: `context.File.GetDiagnostics()` is the only source of Proj4000/4001/4002, but none of the new INI nodes override `GetDiagnostics` and `IniFileParser` doesn't emit an unparsable/error marker, so this always comes back empty. The old `IniFileSyntax`/`HeaderSyntax`/`KeyValuePairSyntax` produced them. `Rules/INI/INI_syntax.cs` still asserts six, so as-is that spec fails (and hangs first on the `[No Header` fixture line, per the other comment). Was dropping the INI syntax rules intended, or should the parser still flag unparsable regions?

src/DotNetProjectFile.Analyzers/Ini/IniFileParser.cs
```diff
@@ -0,0 +1,149 @@
...
+    private static readonly Lexer end = eol | eos;
+    private static readonly Lexer comment = (line_comment("#") | line_comment(";")).optional;
+
+    private static readonly Lexer colon = ch(':', Kind.ColonToken);
+    private static readonly Lexer equals = ch('=', Kind.EqualsToken);
+    private static readonly Lexer assign = equals | colon;
+
+    private static readonly Lexer header = reg(@"^\[[^[\]]+\]", Kind.HeaderToken);
+    private static readonly Lexer key = reg(@"^[a-zA-Z0-9_\-\.]+", Kind.KeyToken);
+    private static readonly Lexer value = reg(@"^[^=:^\s#;][^\s#;]*", Kind.HeaderToken);
```
JohnVerheij: Should this be `Kind.ValueToken`? As-is the value token is tagged `HeaderToken`, and `IniValue.Text` looks for `ValueToken`, so every value's `Text` comes back null and `IniFile.IsRoot` never sees `root = true` (editorconfig root detection stops working).

```suggestion
    private static readonly Lexer value = reg(@"^[^=:^\s#;][^\s#;]*", Kind.ValueToken);
```

JohnVerheij (commented):
Went through the whole diff (Grammr core, the INI/Glob/Git parsers, the node model, the consumers). The direction is good. `SourceReader` + append-only `TokenStream` is much easier to follow than the old `Syntax`/`ResultQueue` stack, and it builds clean.

Two things I think block correctness right now, both inline:

- `IniFileParser.Parse` loops forever on malformed input.
- `Proj4000/4001/4002` can't fire anymore, so `Rules/INI/INI_syntax` still expects six diagnostics and now gets none.

A few behaviors the old stack guaranteed that the rewrite drops. Smaller, but worth keeping:

- `IOFile.SourceText()` opens via the throwing `OpenRead()`, so a locked or permission-denied props file throws `IOException` past the `XmlException` catch in `MsBuildProject.Load` (AD0001). `TryOpenText()` used to swallow that. It's the v1.5.1 Proj3000 "inaccessible files shouldn't crash" case.
- `Grammr/Lexers/Reg` dropped the old 10s regex `Timeout` and turned on `RegexOptions.Compiled` (Reflection.Emit in the VS host). Nothing has super-linear backtracking today, so it's just the safety net that's gone.
- `Grammr/TokenStream` seeds every stream from a shared static `Empty` array and `Add` writes into it in place. With `EnableConcurrentExecution` two files parsing at once can both write `Empty[0]`. A zero-length sentinel would force the first `Add` to allocate per instance.
- `IO/GlobParser.Literal` does `stackalloc char[reader.Length]` sized to the whole remaining file, and the gitignore path hands it the whole file, so a large `.gitignore` can StackOverflow.

None of this is about the approach itself, which I like. Two latent ones I didn't inline (no caller today): `IniSection` token spans overlap on multi-section files (`[a]\n[b]\n`), and a zero-token node throws on `TextSpan`/`FullSpan` where the old `Node` had an empty guard. Happy to send a failing INI spec for the loop if that helps.

src/DotNetProjectFile.Analyzers/Analyzers/Ini/IniSyntaxAnalyzer.cs
Corniel: It still has to be re-implemented. This PR is already huge. The rules are currently disabled by default, because there were some heavy issues with it.

src/DotNetProjectFile.Analyzers/Ini/IniFileParser.cs
Corniel: Yeah, I thought scenario's like these too. When driving home today I already thought that a line starting with `ws + '['`, should always match a header (and report on missing header text or closing header token).

I'm not sure yet which of the other cases should match as `IniEntry` or as a token that is *just* unparsable.

Corniel:
> Went through the whole diff (Grammr core, the INI/Glob/Git parsers, the node model, the consumers). The direction is good. `SourceReader` + append-only `TokenStream` is much easier to follow than the old `Syntax`/`ResultQueue` stack, and it builds clean.

The insight to use that `ref struct SourceReader` was one of my better ones. The previous one already had the append only `TokenStream` but that was a bit obfuscated by the rest. In my first attempt, I tried to combine both lexers and parsers using the `+` operator, hoping to also make parsers look like formal grammar. By dropping that idea, the code became way cleaner, and faster.

> A few behaviors the old stack guaranteed that the rewrite drops. Smaller, but worth keeping:
> 
>     * `IOFile.SourceText()` opens via the throwing `OpenRead()`, so a locked or permission-denied props file throws `IOException` past the `XmlException` catch in `MsBuildProject.Load` (AD0001). `TryOpenText()` used to swallow that. It's the v1.5.1 Proj3000 "inaccessible files shouldn't crash" case.

Well spotted. I re-enabled it.

>     * `Grammr/Lexers/Reg` dropped the old 10s regex `Timeout` and turned on `RegexOptions.Compiled` (Reflection.Emit in the VS host). Nothing has super-linear backtracking today, so it's just the safety net that's gone.

Also re-enabled. To keep things clean I had a separate project/solution first to get everything working outside this solution. All example implementations where defined in .NET 10, which allows the use of  `[GeneratedRegex]`. I just forgot to focus on the fact that is not available in .NET Standard 2.0, and therefor paid less attention to this path.

>     * `Grammr/TokenStream` seeds every stream from a shared static `Empty` array and `Add` writes into it in place. With `EnableConcurrentExecution` two files parsing at once can both write `Empty[0]`. A zero-length sentinel would force the first `Add` to allocate per instance.

I updated the implementation.

>     * `IO/GlobParser.Literal` does `stackalloc char[reader.Length]` sized to the whole remaining file, and the gitignore path hands it the whole file, so a large `.gitignore` can StackOverflow.

I have to think about this. This could indeed be an issue. The Glob parser also does not support ranges like `[a-z]` yet, and I think I can improve handling also by applying the matching on `IOFile` and `IODirectory` instead of `string`. But not in this PR.

> None of this is about the approach itself, which I like. Two latent ones I didn't inline (no caller today): `IniSection` token spans overlap on multi-section files (`[a]\n[b]\n`), and a zero-token node throws on `TextSpan`/`FullSpan` where the old `Node` had an empty guard. Happy to send a failing INI spec for the loop if that helps.

If you have an failing INI spec, that always helps.

JohnVerheij:
Here's a failing spec. It only checks that each line terminates (background-thread watchdog, since `[Timeout]` won't interrupt a tight loop), so it stays out of the `IniEntry`-vs-unparsable question and leaves that to you. Fails 5/5 on the current head, passes once the `else` makes progress. Probably worth landing together with the fix, since each failing case spins a thread until the run ends.

```csharp
using DotNetProjectFile.Ini;
using DotNetProjectFile.IO;
using Grammr;
using Microsoft.CodeAnalysis.Text;
using System.Threading;

namespace Specs.INI.IniFile_terminates_specs;

public class Terminates
{
    [TestCase("@\n")]
    [TestCase("foo bar\n")]
    [TestCase("= value\n")]
    [TestCase("root\n")]
    [TestCase("[No Header\n")]
    public void On_unparsable_line(string content)
    {
        var tree = new GrammrTree(IOFile.Parse("malformed.ini"), SourceText.From(content));

        Exception? error = null;
        var thread = new Thread(() =>
        {
            try { IniFile.Parse(tree); }
            catch (Exception e) { error = e; }
        })
        { IsBackground = true };
        thread.Start();

        thread.Join(TimeSpan.FromSeconds(5)).Should()
            .BeTrue("parsing must make progress on every line, even unparsable ones");
        error.Should().BeNull();
    }
}
```

Corniel:
@JohnVerheij I've solved the issues, and re-implemented GetDiagnostics() on the `GrammrNode`. Therefor, now one test fails, as I still have to think of a better way to report on the right invalid token (mainly: what to do, if there is no next token to report on). I'll do that tomorrow. I think I addressed all your other concerns.

PR #847, opened 2026-07-10
Use AnalyzerType metadata on AdditionalFiles

Corniel:
By adding metadata (`AnalyzerType`) to the `<AdditionalFiles>` it should become easier to assign the right analyzers to the files to analyze. Ideally, we should not have to check the file system anymore.

### DIFF
--- a/specs/Specs/Rules/MS_Build/Build_actions_should_have_single_task.cs
+++ b/specs/Specs/Rules/MS_Build/Build_actions_should_have_single_task.cs
@@ -30,7 +30,7 @@ public class Reports
 
           <ItemGroup>
             <GlobalAnalyzerConfigFiles
-              Include="$([MSBuild]::GetPathOfFileAbove('.globalconfig')"
+              Include="$([MSBuild]::GetPathOfFileAbove('.globalconfig'))"
               Exclude="$(GlobalAnalyzerConfigFiles)" />
           </ItemGroup>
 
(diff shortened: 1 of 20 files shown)


specs/Specs/Rules/MS_Build/Build_actions_should_have_single_task.cs
```diff
@@ -30,7 +30,7 @@ public class Reports
 
           <ItemGroup>
             <GlobalAnalyzerConfigFiles
-              Include="$([MSBuild]::GetPathOfFileAbove('.globalconfig', '$(MSBuildThisFileDirectory)'))"
+              Include="$([MSBuild]::GetPathOfFileAbove('.globalconfig')"
```
JohnVerheij: Missing the closing paren on the `$(…)`. Harmless for what `multiple_tasks` asserts, but worth fixing while you're tidying these `GetPathOfFileAbove` calls.

```suggestion
              Include="$([MSBuild]::GetPathOfFileAbove('.globalconfig'))"
```

JohnVerheij (commented):
Direction looks right. Declaring `AnalyzerType` and routing on it beats sniffing paths/extensions, and compiler-visible item metadata is the right hook for the analyzer side to read later.

Pulled the branch and exercised the SDK path: dropped an `obj/leaked.props` under the `.net.csproj` project and confirmed the `Remove` block in `AdditionalFiles.Sdk.props` still strips it, so the recursive `**/*.props` / `**/*.targets` / `**/*.??proj` globs stay clean.

PR #851, opened 2026-07-15
 Primary select analyzers based on the AnalyzerType

Corniel:
With #847 it became possible to detect analyzers based on metadata. By reading <AdditionalFiles AnalyzerType="{type}"> are now selected. They still have a fallback on an `IOFile` predicate.


src/DotNetProjectFile.Analyzers/Analyzers/AnalyzerRegistry.cs
```diff
@@ -65,32 +51,14 @@ public void RegisterNuGetConfigFileAction(Action<NuGetConfigFileAnalysisContext>
...
+            => context.RegisterAdditionalFileAction(c =>
             {
                 if (ProjectFiles.Global.UpdateResourceFile(c) is { IsXml: true } resource)
                 {
                     register(new(resource, c.Compilation, c.Options, c.CancellationToken, c.ReportDiagnostic));
                 }
             });
 
-            // Fallback for detecting files that not have been added as additional files.
-            context.RegisterCompilationAction(c =>
```
JohnVerheij, line 78: This fallback was the only path where `Resx.AddAdditionalFile` could fire: via `RegisterAdditionalFileAction` the file is by definition additional, so `!IsAdditional` never holds. Removing it makes Proj0006 (resx) unreachable and fails its Reports spec, plus `Sort_data_alphabetically.Reports` (ResxUnsorted.csproj declares no AdditionalFiles).

src/DotNetProjectFile.Analyzers/build/AdditionalFiles.Sdk.props
```diff
@@ -10,7 +10,7 @@
 
     <!-- .editorconfig -->
     <AdditionalFiles
-      Include="$([MSBuild]::GetPathOfFileAbove('.editorconfig'))"
+      Include="$([MSBuild]::GetPathOfFileAbove('.editorconfig', $(MSBuildProjectDirectory)))"
```
JohnVerheij, line 13: `AnalyzerType="SDK"` a few lines up has no matching enum member, so it parses to `None` and the .net.csproj only gets picked up through the `Is.MsBuild` fallback. Add `SDK` to the enum, or use `MSBuildProject` here?

src/DotNetProjectFile.Analyzers/Extensions/Microsoft.CodeAnalysis.AdditionalFileAnalysisContext.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        public AnalyzerType AnalyzerType
+            => context.Options.AnalyzerConfigOptionsProvider
+                .GetOptions(context.AdditionalFile)
+                .TryGetValue("build_metadata.AdditionalFiles.AnalyzerType", out var value)
+            && Enum.TryParse<AnalyzerType>(value, ignoreCase: true, out var type)
+                ? type
+                : AnalyzerType.None;
+
+        /// <summary>Applies if the analyzer type matches or if they not disagree and the predicate is true.</summary>
```
JohnVerheij: "or if they not disagree" reads odd, maybe "or if no analyzer type is set and the predicate is true".

JohnVerheij (commented):
Nice to see the Proj4000 series coming back. A few things I ran into:

The two RESX Reports specs rely on the removed fallback: ResxUnsorted.csproj has no AdditionalFiles, so Resources.resx no longer reaches the analyzers (verified locally: `project_files_not_additional` and `unsorted_data` are the two failing specs on this branch). For Proj2002 the fixture can declare the resx. Proj0006 is more fundamental: the resx analyzer now only runs on files that already are additional files, so `!IsAdditional` is always false and the rule can never fire. If the `**/*.resx` include in AdditionalFiles.targets makes the resx flavor obsolete, should it be deprecated instead of left unreachable? The MS Build flavor is fine, that fallback is still in place.

DotNetProjectFile.Analyzers.props still has `GetPathOfFileAbove('.globalconfig')` without a starting directory. From the NuGet cache that resolves relative to the package folder, the same issue this PR fixes for .editorconfig and NuGet.config.

One scope question: the removed fallback walked all ancestor .editorconfig files until `root = true`, for every project. Now only the nearest one above the .net.csproj is analyzed, and projects without a .net.csproj get no editorconfig analysis unless they add the file as additional themselves. Is that the intended scope for enabling Proj4000-Proj4051 by default? Might be worth a line in the release notes.

src/DotNetProjectFile.Analyzers/Analyzers/AnalyzerRegistry.cs
Corniel: As the analyzers now include `**/*.resx` I see no reason to keep this mechanisme alive. As a result, Proj0006 should be deprecated I guess.

JohnVerheij: Deprecating it sounds right, the `**/*.resx` include covers this by design now. The other failing spec (`Sort_data_alphabetically.Reports`) just needs `ResxUnsorted.csproj` to declare the resx as an additional file.

Corniel: At least, for detecting RESX files. I've removed that code and those tests. The only reason an RESX file is now not an addtional file is when it explicitly removed from the additional files, or if the RESX is not somewhere in the child directories. The first we should not report on, and the second we already did not catch anyway.

src/DotNetProjectFile.Analyzers/build/AdditionalFiles.Sdk.props
Corniel: It should be `SDK`, and I have adjusted it accordingly: rules that apply to `SDK` and `MSBuildProject` are quite different by nature, and having two values makes that distinction clearly.

src/DotNetProjectFile.Analyzers/Extensions/Microsoft.CodeAnalysis.AdditionalFileAnalysisContext.cs
Corniel: I've adjusted the summary.

Corniel:
> The two RESX Reports specs rely on the removed fallback: ResxUnsorted.csproj has no AdditionalFiles, so Resources.resx no longer reaches the analyzers (verified locally: `project_files_not_additional` and `unsorted_data` are the two failing specs on this branch).

I now have updated these project files to explicitly add the RESX-es. They succeed again.

> DotNetProjectFile.Analyzers.props still has `GetPathOfFileAbove('.globalconfig')` without a starting directory. From the NuGet cache that resolves relative to the package folder, the same issue this PR fixes for .editorconfig and NuGet.config.

I've updated the statement. I just missed it before. In a follow-up PR I'll add at least one globalconfig specific rule.

> One scope question: the removed fallback walked all ancestor .editorconfig files until `root = true`, for every project. Now only the nearest one above the .net.csproj is analyzed, and projects without a .net.csproj get no editorconfig analysis unless they add the file as additional themselves. Is that the intended scope for enabling Proj4000-Proj4051 by default? Might be worth a line in the release notes.

I have to investigate how to handle those scenario's. Feel free to create an issue for this.

I think I addressed all your concerns, can you have a look again @JohnVerheij?

JohnVerheij (approved):
All good as far as I can tell. Verified locally on 3b15bef: the RESX, EditorConfig, SDK, INI and Add_additional_files specs all pass, and the AdditionalFileItemNames approach is a nicer fix than declaring the resx explicitly.

PR #852, opened 2026-07-15
Proj0052: Do not advise attributes in child nodes of the root

Corniel:
When testing a project file (for other reasons), I stumbled on the following:

``` xml

  <PropertyGroup>
  </PropertyGroup>

  <ItemGroup>
    <!--<AdditionalFiles Include=".editorconfig" AnalyzerType="EditorConfig" />-->
  </ItemGroup>

</Project>
```
The analyzers reported: [Proj0052: Consider <ItemGroup> as an attribute](https://dotnet-project-file-analyzers.github.io/rules/Proj0052.html). (It also correctly reprots that `<ItemGroup>` is empty and can be removed).


src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/PreferAttributesOverElements.cs
```diff
@@ -28,6 +28,7 @@ is not PropertyGroup
         and not ItemDefinitionGroup;
 
     private static bool CanBeAttribute(Node node)
-        => node.Element.Attributes().None()
+        => node.Depth >= 2
```
JohnVerheij, line 31: The depth guard is right for root children. A node that only contains comments is still a leaf with an empty Value though, so inside a Target this still reports: `<Target Name="T"><ItemGroup><!-- todo --></ItemGroup></Target>` gives "Consider ItemGroup as an attribute", and applying the advice would drop the comment. Skipping nodes that have comment children in CanBeAttribute would close that; also fine as a follow-up.

JohnVerheij (approved):
Fix verified on the branch: the root-level ItemGroup case no longer reports and the existing expectations hold. One note on an adjacent case.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/PreferAttributesOverElements.cs
Corniel: That is a nice spot. I added another case (a `<ItemGroup>` within a `<Choose>`) that should not trigger.

I'm not sure if I would like to skip nodes contain a comment though. Why would you like that?

Corniel: @JohnVerheij I'll merge this PR. If you think want nodes with comments to be excluded, please open an issue and share your reasoning.

JohnVerheij: Reasoning in #857.

Issue #854, opened 2026-07-15
dotnet_diagnostic severity entries belong in a globalconfig


Corniel:
It is possible to define

 ``` ini
dotnet_diagnostic.CA1861.severity = none
```

in an `.editorconfig` file. However, some rules (inlcuding [SYSLIB1045](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1040-1049) can only be configured in `.globalconfig` files.

Besides that, having diagnostic configuration in a separate file from the file where formatting that is file type specific is defined is a good separation of concerns.

Frulfump:
It makes sense for some rules that are required like you mentioned. The benefit of separation of concerns doesn't outweigh the usefulness of being able to set different severities at the file pattern level which is only possible when doing it in the .editorconfig file. So ideally I'd like to see a rule that makes you aware of the fact that a specific rule must be set in globalconfig to take effect.

Setting different severities for test projects is not that uncommon.

PR #855, opened 2026-07-15
Proj4025: Configure static code analysis in a .globalconfig file

### LINKED ISSUE #407: New Rule: Consider globalconfig
When people use `<NoWarn>` or `<WarningsError>` it might be a good idea to suggest (default severity should be info) to use a `.globalconfig` file instead.

Corniel:
No global analyzer configuration file (`.globalconfig`) was resolved during the project evaluation phase. 

Using a global configuration via `.globalconfig` allows to centrally define which rules are part of the [definition of done](https://en.wikipedia.org/wiki/Scrum_%28project_management%29#Definition_of_Done_%28DoD%29) (or whatever code quality gate the team agreed on). Furthermore, by separating this configuration from the `.editconfig` is considerd a good seperation of
concerns.

Using a global configuration via `.globalconfig` allows a team to centrally define which static analysis rules are part of their [Definition of Done](https://en.wikipedia.org/wiki/Scrum_%28project_management%29#Definition_of_Done_%28DoD%29) (or whichever quality gate the team has agreed upon). 

Furthermore, separating this compiler-level diagnostic configuration from the IDE-focused `.editorconfig` file is considered a good [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). While `.editorconfig` excels at formatting rules (tabs, spaces, encoding, line-endings), `.globalconfig` is built purely for compiler and analyzer behavior.

There are also edge cases where certain diagnostic severities defined in an `.editorconfig` (such as [SYSLIB1045](https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib1040-1049)) are silently ignored by the compiler, but are correctly applied when defined inside a `.globalconfig`.

## Compliant
Place a file named `.globalconfig` in the root directory of your repository. Ensure it contains the following entry at the top:
```ini
is_global = true
```
or, if your global configuration file uses a custom name or resides in a specific folder, explicitly add it to your project or `Directory.Build.props` file using the `GlobalAnalyzerConfigFiles` item group:

```xml
<ItemGroup>
  <GlobalAnalyzerConfigFiles Include="{some path}" />
</ItemGroup>
```

Closes #407


Corniel:
I still have to add a test for this, and I think I know how to make this work. :)

Frulfump:
I don't think we will be able to change the severity of this rule itself in globalconfig which is an issue for me as I think the rule is misguided and will want to disable it asap. My other concern about the feature itself is in the linked issue.

src/DotNetProjectFile.Analyzers/build/DotNetProjectFile.Analyzers.targets
```diff
@@ -63,12 +63,20 @@
...
     <Warning
       Condition="'$(IsRoslynBased)' != 'true' and $([MSBuild]::GetPathOfFileAbove('.net.csproj', '$(MSBuildThisFileDirectory)')) == ''"
       Code="Proj1007"
       Text="Use .net.csproj to analyze this project"
       HelpLink="https://dotnet-project-file-analyzers.github.io/rules/Proj1007.html" />
   </Target>
 
+  <Target Name="ValidateGlobalAnalyzerConfigFiles" AfterTargets="PrepareForBuild;ResolveProjectReferences">
+    <Warning
+      Condition="'@(GlobalAnalyzerConfigFiles)' == ''"
```
JohnVerheij:
```suggestion
      Condition="'@(GlobalAnalyzerConfigFiles->Exists())' == ''"
```

docs/rules/Proj4025.md
```diff
@@ -0,0 +1,45 @@
+---
+parent: INI
+ancestor: Rules
+permalink: /rules/Proj4025
+---
+
+# Proj4025: Configure static code analysis in a .globalconfig file
+No global analyzer configuration file (`.globalconfig`) was resolved during the
```
Frulfump, line 8: So you need to have the file even if you have no rules in it or no rules in your editorconfig which you could move there? If so sounds annoying having to add an empty file for no reason.

docs/rules/Proj4025.md
```diff
@@ -0,0 +1,45 @@
...
+project evaluation phase. 
+
+Using a global configuration via `.globalconfig` allows to centrally define
+which rules are part of the [definition of done](https://en.wikipedia.org/wiki/Scrum_%28project_management%29#Definition_of_Done_%28DoD%29)
+(or whatever code quality gate the team agreed on). Furthermore, by separating
+this configuration from the `.editconfig` is considerd a good seperation of
+concerns.
```
JohnVerheij, line 15: This paragraph is the earlier draft of the two below it; can be removed (it also carries the `.editconfig`/"considerd"/"seperation" typos).

docs/rules/Proj4025.md
```diff
@@ -0,0 +1,45 @@
...
+
+Using a global configuration via `.globalconfig` allows a team to centrally
```
Frulfump, line 17: You can already do this today with an editorconfig. Also worth noting that editorconfig will take precedence so globalconfig can't enforce what it says.

JohnVerheij (commented):
The warning can never fire on regular C# and VB projects: Roslyn's Microsoft.Managed.Core.targets speculatively adds a `.globalconfig` path for every directory above the project to `GlobalAnalyzerConfigFiles`, whether the file exists or not (the compiler filters to existing files later, using `->Exists()`). The list is built from the ancestors of the `Compile` items, so it is only empty for projects without compile items, such as a .net.csproj proxy. Verified with a bare net10 console app with no `.globalconfig` anywhere: the list contains 12 speculative ancestor paths and no warning is raised.

Filtering the same way Roslyn does fixes it for both cases, verified both ways (fires on the bare project, silent once a `.globalconfig` exists): see the inline suggestion.

The doc page also has the intro twice (the "centrally define / Definition of Done" paragraph appears in two variants).

docs/rules/Proj4025.md
```diff
@@ -0,0 +1,45 @@
...
+---
+
+# Proj4025: Configure static code analysis in a .globalconfig file
+No global analyzer configuration file (`.globalconfig`) was resolved during the
+project evaluation phase. 
+
+Using a global configuration via `.globalconfig` allows to centrally define
+which rules are part of the [definition of done](https://en.wikipedia.org/wiki/Scrum_%28project_management%29#Definition_of_Done_%28DoD%29)
+(or whatever code quality gate the team agreed on). Furthermore, by separating
+this configuration from the `.editconfig` is considerd a good seperation of
```
Frulfump, line 14: You can just have a separate section in the editorconfig file and I see no major projects doing this type of separation intentionally.

docs/rules/Proj4025.md
```diff
@@ -0,0 +1,45 @@
...
+Using a global configuration via `.globalconfig` allows a team to centrally
+define which static analysis rules are part of their [Definition of Done](https://en.wikipedia.org/wiki/Scrum_%28project_management%29#Definition_of_Done_%28DoD%29)
+(or whichever quality gate the team has agreed upon). 
+
+Furthermore, separating this compiler-level diagnostic configuration from the
+IDE-focused `.editorconfig` file is considered a good [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns).
+While `.editorconfig` excels at formatting rules (tabs, spaces, encoding, line-endings),
+`.globalconfig` is built purely for compiler and analyzer behavior.
+
+There are also edge cases where certain diagnostic severities defined in an
```
Frulfump, line 26: Would be better to just find those and warn to move them to a globalconfig file

Frulfump (commented):
See more feature concerns in the issue itself. Hoping that the discussion takes place there as the PR review should focus on the implementation itself.

docs/rules/Proj4025.md
```diff
@@ -0,0 +1,45 @@
...
+`.editorconfig` (such as [SYSLIB1045](https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib1040-1049))
+are silently ignored by the compiler, but are correctly applied when defined
+inside a `.globalconfig`.
+
+## Compliant
+Place a file named `.globalconfig` in the root directory of your repository.
+Ensure it contains the following entry at the top:
+```ini
+is_global = true
```
Frulfump, line 35: This isn't strictly required when the file i named exactly globalconfig but it's recommended.

docs/rules/Proj4025.md
Corniel: That is rule that is already planned: #854.

Corniel:
> I don't think we will be able to change the severity of this rule itself in globalconfig which is an issue for me as I think the rule is misguided and will want to disable it asap. My other concern about the feature itself is in the linked issue.

That is fair point. I just had a walk and came with a (general applicable) solution: lets MSBuild assign a property based on this condition, and let an analyzer report based on the value of the property.

Corniel:
I've decided to drop this one: we now have a rule that advises to move things from the `.editorconfig` to `.globalconfig`, that work beter than this general rule.

PR #858, opened 2026-07-16
Extend lexers to both work on SourceReader as ReadOnly<char>

Corniel:
To be able to use `Lexer`'s both within parsers (using `SourceReader`) as while matching GLOB's, only dealing with `ReadOnlySpan<char>`.
- [x] Extend contract
- [x] Extend implementations
- [x] Alias `ReadOnlySpan<char>` as `Chars`
- [x] Implement `CharRange` for matching characters in a certain range.

### DIFF
--- a/specs/Specs/AwesomeAssertions/LexerAssertions.cs
+++ b/specs/Specs/AwesomeAssertions/LexerAssertions.cs
@@ -1,4 +1,5 @@
 using AwesomeAssertions.Execution;
+using Grammr;
 using Grammr.Lexers;
 using System.Diagnostics;
 
@@ -13,7 +14,7 @@ internal sealed class LexerAssertions(Lexer subject, AssertionChain? chain = nul
     [DebuggerStepThrough]
     public AndConstraint<LexerAssertions> Match(string match, string from, string because = "")
     {
-        var length = Subject.Match(new(from));
+        var length = Subject.Match(new SourceReader(match));
 
         Assert
             .ForCondition(length is { })
@@ -29,7 +30,7 @@ internal sealed class LexerAssertions(Lexer subject, AssertionChain? chain = nul
     [DebuggerStepThrough]
     public AndConstraint<LexerAssertions> NotMatch(string source, string because = "")
     {
-        var length = Subject.Match(new(source));
+        var length = Subject.Match(new SourceReader(source));
 
         Assert
             .ForCondition(length is null)
--- a/specs/Specs/AwesomeAssertions/LexerAssertions.cs
+++ b/specs/Specs/AwesomeAssertions/LexerAssertions.cs
@@ -14,7 +14,7 @@ internal sealed class LexerAssertions(Lexer subject, AssertionChain? chain = nul
     [DebuggerStepThrough]
     public AndConstraint<LexerAssertions> Match(string match, string from, string because = "")
     {
-        var length = Subject.Match(new SourceReader(match));
+        var length = Subject.Match(new SourceReader(from));
 
         Assert
             .ForCondition(length is { })
--- a/src/DotNetProjectFile.Analyzers/Grammr/Lexers/Reg.cs
+++ b/src/DotNetProjectFile.Analyzers/Grammr/Lexers/Reg.cs
@@ -18,6 +18,11 @@ internal sealed class Reg(Regex pattern, string? kind) : Lexer(kind)
 
     /// <inheritdoc />
     [Pure]
+    public override int? Match(Chars span)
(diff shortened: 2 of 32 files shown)


CptWesley (approved):
LGTM

specs/Specs/AwesomeAssertions/LexerAssertions.cs
```diff
@@ -13,7 +14,7 @@ internal sealed class LexerAssertions(Lexer subject, AssertionChain? chain = nul
     [DebuggerStepThrough]
     public AndConstraint<LexerAssertions> Match(string match, string from, string because = "")
     {
-        var length = Subject.Match(new(from));
+        var length = Subject.Match(new SourceReader(match));
```
JohnVerheij: This now lexes `match` instead of `from`, so the prefix cases no longer exercise the lexer against trailing input. `String.Parses("grammr as first word", "grammr")` only ever feeds `"grammr"` to the lexer; `from` is just used for slicing afterwards.

It also lets incorrect lexers pass. This passes on this branch, but should fail:

```csharp
match(IsDigit).Should().Match("12", from: "123");
```

With `from` it fails with "matched 123 instead", as intended.

```suggestion
        var length = Subject.Match(new SourceReader(from));
```

src/DotNetProjectFile.Analyzers/Grammr/Lexers/_shared.cs
```diff
@@ -87,6 +87,28 @@ public static class Shared
...
+    /// </remarks>
+    /// <returns>
+    /// A new lexer.
+    /// </returns>
+    public static Lexer range(string range, [CallerMemberName] string? kind = null) => range.Length switch
+    {
+        3 when range[0] < range[2] && range[1] is '-'
+            => new CharRange(range[0], range[2], kind),
+
+        _ => throw new ArgumentOutOfRangeException(nameof(range), "Length must be two with the first character be before the second value."),
```
JohnVerheij, line 109: The message says the length must be two, but the switch requires three (`"x-z"`). Something like `The range must be formatted like "a-z", with the first character before the second.` covers both conditions.

Two related nits:
- the remark above (lines 98-99) says the second character has no meaning, but `range[1] is '-'` is validated now, so it does.
- `range[0] < range[2]` rejects `"a-a"`. `char.IsBetween` allows equal bounds; is excluding the single-character range intended?

src/DotNetProjectFile.Analyzers/Grammr/Lexers/Reg.cs
```diff
@@ -16,6 +16,11 @@ public Reg(string pattern, string? kind) : this(new Regex(pattern, Options, Time
 
     private readonly Regex Pattern = pattern;
 
+    /// <inheritdoc />
+    [Pure]
+    public override int? Match(Chars span)
+        => throw new NotSupportedException("Regex is not supported as it requires a lot of string allocations.");
```
JohnVerheij, line 22: A `reg` nested in a `Choice` or `Optional` works on the reader path but throws here on the span path, so the same grammar behaves differently depending on the entry point. Nothing hits that today, but once GLOB matching starts feeding grammars through `Match(Chars)` this only surfaces at runtime. Worth a note on `Lexer.Match(Chars)` that span-fed grammars must be regex-free, or is the plan to keep `reg` out of those grammars entirely?

JohnVerheij (commented):
The dual-path setup looks good: `CharRange` checks out including the unsigned wrap-around for chars below the minimum, and `Choice`/`Optional` keep both paths symmetric. One real issue in `LexerAssertions` (inline), plus two small ones.

src/DotNetProjectFile.Analyzers/Grammr/Lexers/_shared.cs
Corniel: Yes. This lexer is for *our* parses, with fixed ranges. Instead of `range("a-a")`, just use `ch('a')`. When parsing `.gitignore`, `[a-a]` wil be allowed, but also flagged as being verbose.

src/DotNetProjectFile.Analyzers/Grammr/Lexers/Reg.cs
Corniel: That is intended. There is no allocation free solution (at least not for .NET standard) that would work. So by not having to create a `SourceText`, and keeping both the `Chars` and the `string` in the `SourceReader` but only relying on the `Chars`, you can not benefit from allocation free `reg`. In most of those cases the constraints can be defined differently anyway.

PR #869, opened 2026-07-22
Blazor with scoped CSS should compile

### LINKED ISSUE #861: Build error with Blazor and scoped CSS files
Hi!
First of all, thanks for the awesome analyzer!

I faced an issue today when upgrading `DotNetProjectFile.Analyzers` from version `1.15.0` to version `1.15.1` in a Blazor project.
In the Blazor project, I'm using scoped CSS files. In case you are not familiar with scoped CSS files, they are just plain CSS files that have a name that must match existing `.razor` files. It is not allowed to have two scoped CSS files with identical name.

The project builds fine with version `1.15.0` of the analyzer. When the only change I make in the repo is to bump the analyzer to version `1.15.1`, I get the following error on my build server:
```
##[error]Features/Components/CompareConditionTreeView.razor(0,0): Error BLAZOR101: More than one scoped css files were found for the razor component 'Features/Components/CompareConditionTreeView.razor'. Each razor component must have at most a single associated scoped css file.
Features/Components/CompareConditionTreeView.razor.css
Features/Components/CompareConditionTreeView.razor.css
Features/Components/CompareConditionTreeView.razor : error BLAZOR101: More than one scoped css files were found for the razor component 'Features/Components/CompareConditionTreeView.razor'. Each razor component must have at most a single associated scoped css file.
```
This might be an issue specific to Linux, as I only see the issue on my Linux build server. Note that the same issue appears with version `1.15.2` as well, but the first version with the issue is version `1.15.1`.

Corniel:
The root cause it that `*.razor.css` files (being `<None>`) are updated to being `<Content>` to support SonarIntegration, but that interferes with the Razor build. 

The best way to fix this would be if Sonar would support reporting on [`<AdditionFiles>`](https://community.sonarsource.com/t/suport-roslyn-analyzers-on-additional-files/123957). For the time being, `**/*.razor.css` is excluded.

Closes #861


Gemberkoekje (approved):
Should this be mentioned in the release notes anywhere?

PR #874, opened 2026-07-24
Remove section header from .globalconfig

### LINKED ISSUE #853: Globalconfig files should only have one section without an header
Globalconfig files should not contain `[..]` header sections.

Corniel:
Closes #853

### DIFF
--- a/docs/rules/Proj4025.md
+++ b/docs/rules/Proj4025.md
@@ -1,0 +1,31 @@
+---
+parent: Globalconfig
+ancestor: Rules
+permalink: /rules/Proj4025
+---
+
+# Proj4025: Remove section header
+Unlike `.editorconfig` files, which use headers (like `[*.cs]`) to scope rules
+to specific files, `.globalconfig` files apply globally. Sections headers have
+no meaning, and should be removed.
+
+## Non-compliant
+``` ini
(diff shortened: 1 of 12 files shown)


docs/rules/Proj4025.md
```diff
@@ -0,0 +1,31 @@
...
+no meaning, and should be removed.
+
+## Non-compliant
+``` ini
+is_global = true
+        
+dotnet_diagnostic.CA1860.severity  = none       #  Prefer comparing 'Length' to 0 rather than using 'Any()'
+dotnet_diagnostic.CS1591.severity  = suggestion # Missing XML comment for publicly visible type or member
+
+[IDE001]
```
Frulfump: Text above uses *.cs instead and I've never seen anyone use [IDExxx] as a header and this rule is mostly geared towards people accidentally including headers you would see in editorconfig so it makes sense to switch the example to be more believable.

Frulfump (approved):

docs/rules/Proj4025.md
Corniel:
```suggestion
[*.cs]
```

Corniel: I agree. I changed it.

PR #891, opened 2026-07-26
Use known diagnostic IDs

### LINKED ISSUE #879: Configured Projxxxx rule in .globalconfig does not exist
We are in full control of all `Projxxxx` rules out there, so if someone configures a `Projxxxx` rule we know not to exist, we should report on these. We should also take `<NoWarn>` into consideration here.

Corniel:
# Proj4029: Use known diagnostic IDs
This rule ensures that configuration entries target valid diagnostic IDs. As we can only certain about our own rule, only diagnostic IDs with the `Proj` prefix are checked.

## Non-compliant
``` ini
is_global = true
        
dotnet_diagnostic.Proj001.severity = warning
```

## Compliant
``` ini
is_global = true
        
dotnet_diagnostic.Proj1001.severity = warning
```

Closes #879

### DIFF
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="../../props/common.props" />
 
--- a/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
+++ b/src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <Import Project="../../props/common.props" />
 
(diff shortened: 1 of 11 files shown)


Frulfump:
Worth noting that the identifier is chosen arbitrarily so someone could theoretically use the same prefix as us. (Added the same to the issue itself). Doesn't mean we can't add this rule it just means that there's theoretical risk of conflict and that the need to be able to silence the rule is slightly higher.

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
Frulfump, line 1: BOM removal or addition? If removal could we clean up all files from BOM in a new PR?

src/DotNetProjectFile.Analyzers/Rule.Ini.cs
```diff
@@ -83,6 +83,17 @@ public static class Ini
                 "the file uses a custom name instead of .globalconfig.",
             tags: [".globalconfig", "diagnostic", "severity"],
             category: Category.Clarity);
+        public static DiagnosticDescriptor UseKnownDiagnosticIds => New(
+            id: 4029,
+            title: "Use valid diagnostic severity value",
+            message: "diagnostic analyzer rule '{0}' is unknown",
+            description:
+                "The compiler requires standard severity keywords to parse .globalconfig " +
+                "files. Using invalid values typos prevents the rule from being " +
```
Frulfump:
```suggestion
                "files. Using invalid values prevents the rule from being " +
```

Frulfump (approved):

src/DotNetProjectFile.Analyzers/DotNetProjectFile.Analyzers.csproj
Corniel: We have none. Visual Studio sometimes adds a BOM despite our `.editorconfig` configuration. I need to create an issue (or) PR but never got the time to do it.

src/DotNetProjectFile.Analyzers/Rule.Ini.cs
Frulfump, line 1: Oh this is just copied I guess from the other rule, should not reference severity here just the diagnostic ID itself

Frulfump:
```suggestion
                "files. Using invalid values prevents the rule from being " +
```

Issue #890, opened 2026-07-26
Advise on disabling rules


Corniel:
By analyzing (common) Roslyn analyzers, we could detect rules that are
* obsolete
* similar
* conflicting

In all three cases we could advise to disable specific rules in the `.globalconfig`. It requires some recursing analysis of a set of analyzers, and some manual configuration.

See #878

Issue #878, opened 2026-07-25
Rely on not-obsolete rules only


Corniel:
Roslyn rules marked as obsolete should not occur in `.globalconfig` unless set to `none`. We can no from some common analyzers which rules these are, and as rules are unlikely to be revitalized after being marked as `[Obsolete]`, it is unlikely we'll report FP's. FN's can occur, and that is okay.

PR #900, opened 2026-07-30
Introduction of Diagnostics project

Corniel:
This project (and supposed NuGet package) contains tons of information about Roslyn rules. It allows:
1. A rule to be created that reports on relying on obsolete Roslyn rules (#878)
2. A rule to be created that reports on a Roslyn rule that is similar to another rule (#890)
3. A rule to be created that reports on a Roslyn rule that is contradicting to another rule (#890)
4. A rule to be created that reports on Roslyn rules being configured for analyzers not enabled
5. Generating a (static) webpage that allows to search for Roslyn rules
6. Generating a `.globalconfig` file based on used rules, specifying the overrides

Obviously, it creates other opportunities too, but this are things I already thought about.

I'm not certain yet about the namespace of this project/package, and may be some (public) names could be different too.


specs/Specs/Diagnostics/AnalyzersInfo_specs.cs
```diff
@@ -0,0 +1,119 @@
...
+    [Test]
+    public void Contains_rules()
+    {
+        var info = DiagnosticCollection.Embedded();
+        info.Packages.Should().AllSatisfy(p => p.Rules.Should().NotBeEmpty(because: p.Id));
+        info.Count.Should().Be(7354);
+    }
+
+    [Test]
+    public void Contains_pakages()
```
Sjaaky:
```suggestion
    public void Contains_packages()
```

specs/Specs/GlobalConfig_specs.cs
```diff
@@ -0,0 +1,150 @@
...
+using static DotNetProjectFile.Diagnostics.GlobalConfigurationEntry;
+namespace GlobalConfig_specs;
+
+[Explicit]
+public class Define
+{
+    [Test]
+    public void Solution_wide_preferences()
+    {
+        var config = GlobalConfiguration.New(true, DiagnosticCollection.Embedded().FromPackages(
```
Sjaaky, line 13: Is this a unittest which uses fluent syntax to create the .globalconfig?

src/DotNetProjectFile.Diagnostics/IO/TempDirectory.cs
```diff
@@ -0,0 +1,51 @@
+using System.IO;
+
+namespace DotNetProjectFile.Diagnostics.IO;
+
+/// <summary>Represents a directory that lives during the lifetime of its scope.</summary>
+/// <remarks>
+/// Should always been used with a using statement.
```
Sjaaky:
```suggestion
/// Should always be used with a using statement.
```

src/DotNetProjectFile.Diagnostics/IO/TempDirectory.cs
```diff
@@ -0,0 +1,51 @@
...
+///
+/// <code>
+/// using(var directory = new TemporaryDirectory()
```
Sjaaky:
```suggestion
/// using(var directory = new TemporaryDirectory())
```

src/DotNetProjectFile.RuleCatalog/DiagnosticInfo.cs
```diff
@@ -0,0 +1,127 @@
...
+    IComparable<DiagnosticInfo>
+{
+    /// <summary>The (first) version of the diagnostic.</summary>
+    [JsonConverter(typeof(NuGetVersionJsonConverter))]
+    [JsonPropertyName("f")]
+    public NuGetVersion? First { get; init; }
+
+    /// <summary>The (latest) version of the diagnostic.</summary>
+    [JsonConverter(typeof(NuGetVersionJsonConverter))]
+    [JsonPropertyName("v")]
```
Sjaaky, line 27: Does the 'v' come from some external definition? Could it just be "version"?

.gitattributes
```diff
@@ -1,5 +1,5 @@
 # Auto detect text files and perform LF normalization
-* text=auto
+* text=auto working-tree-encoding=UTF-8
```
Sjaaky: Why is "working-tree-encoding=UTF-8" added?
Accoring to the documentation this setting is only needed when the files in the working-tree are not in utf-8.
I'm very suspicious about setting this for all files.
Is this an attempt to automatically strip BOM's?

https://git-scm.com/docs/gitattributes#_working_tree_encoding
"Use the working-tree-encoding attribute only if you cannot store a file in UTF-8 encoding and if you want Git to be able to process the content as text."
"Reencoding content requires resources that might slow down certain Git operations (e.g git checkout or git add)."

AGENTS.md
```diff
@@ -73,3 +74,4 @@ Rule numbering: `Proj00xx` general, `Proj02xx` packaging, `Proj03xx` NuGet confi
 ## Gotchas
 
 * **Do not modify `.net.csproj`:** This file is exclusively for running Roslyn Analyzers. Agents must completely ignore and skip this file during development.
+* **Do not modify or read `src/DotNetProjectFile.Diagnostics/Data/`DiagnosticCollection.json`:** This file is huge, and is data about analyzers.
```
Sjaaky:
```suggestion
* **Do not modify or read `src/DotNetProjectFile.Diagnostics/Data/DiagnosticCollection.json`:** This file is huge, and is data about analyzers.
```

Sjaaky (dismissed):

src/DotNetProjectFile.RuleCatalog/IO/TemporaryDirectory.cs
```diff
@@ -0,0 +1,51 @@
...
+/// <code>
+/// using(var directory = new TemporaryDirectory()
+/// {
+///     // Do stuff.
+/// }
+/// </code>
+/// </remarks>
+internal sealed class TemporaryDirectory : IDisposable
+{
+    /// <summary>Initializes a new instance of the <see cref="TemporaryDirectory" /> class.</summary>
```
Sjaaky, line 18:
```suggestion
    /// <summary>Creates a new temporary directory, which will be deleted when this instance is disposed.</summary>
```

src/DotNetProjectFile.RuleCatalog/DiagnosticInfo.cs
Corniel: It could be `version`, but it is a data structure, with over 3500 entries, that will be embedded.

.gitattributes
Corniel: It is/was such an attempt/experiment indeed. We could consider to only but it on `*.csproj` as this is the file that faces the issue now and then (or try to address this with the Visual Studio team).

specs/Specs/GlobalConfig_specs.cs
Corniel: Yes it is. You think that is not clear enough?

.gitattributes
Corniel:
```suggestion
* text=auto
```

src/DotNetProjectFile.RuleCatalog/DiagnosticId.cs
```diff
@@ -0,0 +1,63 @@
...
+    public string? Prefix
+        => Pattern.Match(value) is { Success: true } match
+        ? match.Groups[nameof(Prefix)].Value
+        : null;
+
+    /// <summary>The (optional) numeric value.</summary>
+    public int? Numeric
+        => Pattern.Match(value) is { Success: true } m && int.TryParse(m.Groups[nameof(Numeric)].Value, out var n)
+        ? n
+        : null;
```
JohnVerheij, line 29: `value` is `null` for `default(DiagnosticId)`, so both `Prefix` and `Numeric` throw, while `ToString()`, `GetHashCode()` and `CompareTo()` all guard for it. Ran it:

```
DiagnosticId.Empty.Prefix     -> ArgumentNullException (Parameter 'input')
DiagnosticId.Empty.Numeric    -> ArgumentNullException
DiagnosticId.Empty.ToString() -> ""
```

For something documented as "Empty/not set" I would expect `null` from both.

src/DotNetProjectFile.RuleCatalog/DiagnosticId.cs
```diff
@@ -0,0 +1,63 @@
...
+        => obj is DiagnosticId other
+        && Equals(other);
+
+    /// <inheritdoc />
+    [Pure]
+    public bool Equals(DiagnosticId other) => value == other.value;
+
+    /// <inheritdoc />
+    [Pure]
+    public int CompareTo(DiagnosticId other) => (value ?? string.Empty).CompareTo(other.value);
```
JohnVerheij, line 51: `string.CompareTo` is `CurrentCulture`, and `DiagnosticCollector` writes `[.. rules.Values.Order()]`, so the order of the committed `DiagnosticCollection.json` depends on the locale of whoever regenerates it. Stable for most cultures, but sorting the 7,437 ids under `da-DK` moves 239 of them relative to the invariant culture, since Danish collates a double `a` as a single letter at the end of the alphabet. On an 83k-line file that is a large phantom diff.

src/DotNetProjectFile.Diagnostics/NugetPackage.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    [Pure]
+    internal NugetPackage Load() => this with { Rules = [.. Rules.Select(r => r.Load(Version))] };
+
+    /// <inheritdoc />
+    [Pure]
+    public int CompareTo(NugetPackage? other)
+        => other is null
+        ? +1
+        : Id.CompareTo(other.Id);
```
JohnVerheij: Same as `DiagnosticId.CompareTo`: `Id.CompareTo` is `CurrentCulture`. `string.CompareOrdinal(Id, other.Id)` keeps the package order locale-independent too.

src/DotNetProjectFile.Diagnostics/DiagnosticInfo.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// </summary>
+    /// <param name="descriptor">
+    /// The matching <see cref="DiagnosticDescriptor"/>.
+    /// </param>
+    [Pure]
+    public static DiagnosticInfo New(DiagnosticDescriptor descriptor) => new()
+    {
+        Id = new(descriptor.Id),
+        Title = descriptor.Title.ToString(),
+        Description = descriptor.Description.ToString(),
```
JohnVerheij: `LocalizableString.ToString()` without a format provider resolves against the ambient UI culture, so a descriptor backed by `LocalizableResourceString` would be captured in the collecting machine's language. `ToString(CultureInfo.InvariantCulture)` on both removes the doubt. Unlike the ordering above, I did not have a localized analyzer to hand to demonstrate this one.

src/DotNetProjectFile.RuleCatalog/DiagnosticInfo.cs
```diff
@@ -0,0 +1,133 @@
...
+    [JsonPropertyName("level")]
+    public DiagnosticSeverity DefaultSeverity { get; init; } = DiagnosticSeverity.Warning;
+
+    /// <inheritdoc cref="DiagnosticDescriptor.IsEnabledByDefault" />
+    [JsonPropertyName("enabled")]
+    public bool IsEnabledByDefault { get; init; } = true;
+
+    /// <summary>The .globalconfig severity.</summary>
+    [JsonIgnore]
+    public GlobalConfigSeverity GlobalConfigSeverity => DefaultSeverity switch
```
JohnVerheij, line 67: This looks like the natural home for the `NotConfigurable` filter from the summary: a rule carrying that tag can never have its severity set from a config, so mapping it to a `GlobalConfigSeverity` at all is what pulls 3,054 unusable entries into scope.

specs/Specs/Specs.csproj
```diff
@@ -42,6 +43,12 @@
     <PackageReference Include="System.Security.Cryptography.Xml" Version="10.0.11" />
     <PackageReference Include="Verify" Version="31.28.0" />
     <PackageReference Include="Verify.NUnit" Version="31.28.0" />
+    <PackageReference Include="NuGet.Common" Version="7.6.0" />
+    <PackageReference Include="NuGet.Frameworks" Version="7.6.0" />
+    <PackageReference Include="NuGet.Packaging" Version="7.6.0" />
+    <PackageReference Include="NUnit" Version="4.6.0" />
+    <PackageReference Include="Verify" Version="31.16.2" />
+    <PackageReference Include="Verify.NUnit" Version="31.16.2" />
```
JohnVerheij: These six duplicate entries already present in the same `Test tools` group directly above, at older versions: `NuGet.Common` 7.9.0, `NuGet.Frameworks` 7.9.0, `NuGet.Packaging` 7.9.0, `NUnit` 4.6.1, `Verify` 31.28.0, `Verify.NUnit` 31.28.0. That is the `NU1504` in the build. Looks like the branch predating the dependabot bumps.

specs/Specs/GlobalConfig_specs.cs
```diff
@@ -0,0 +1,150 @@
...
+            "AsyncFixer",
+            "AwesomeAssertions.Analyzers",
```
JohnVerheij, line 15: The embedded collection has `FluentAssertions.Analyzers` and no `AwesomeAssertions.Analyzers`, and `FromPackages` filters with `Where(p => packageIds.Contains(p.Id))`, so this id yields nothing and says nothing. None of the 109 ids in the generated `.globalconfig` come from any assertions analyzer as a result. The repo does reference `AwesomeAssertions` 9.6.0, so the intent is clear enough; the corpus simply does not carry its analyzer package. An overload that reports unknown ids would have caught it.

.globalconfig
```diff
@@ -1,126 +1,121 @@
...
-dotnet_diagnostic.IDE0130.severity  = none       # Namespace does not match file structure [Justification: We do not care.]
+dotnet_diagnostic.IDE0130.severity  = none       # Namespace does not match folder structure [Justification: We do not care.]
 dotnet_diagnostic.IDE1006.severity  = none       # Naming Styles [Justification: Naming styles should be guarded by other rules.]
 
-dotnet_diagnostic.IDE0001.severity  = warning    # Simplify name
-dotnet_diagnostic.IDE0002.severity  = warning    # Simplify member access
-dotnet_diagnostic.IDE0004.severity  = warning    # Remove unnecessary cast
-dotnet_diagnostic.IDE0005.severity  = warning    # Using directive is unnecessary
-dotnet_diagnostic.IDE0017.severity  = warning    # Use object initializers
+dotnet_diagnostic.IDE0001.severity  = warning
```
JohnVerheij: These three lines end in trailing whitespace. Root cause is in `GlobalConfigurationEntry.ToString()`, noted there.

They are also a live example of capability 4: of the 109 ids in this file, `IDE0001`, `IDE0002` and `IDE0038` are the only three the collection does not know about, and they are exactly the three that come out with a dangling pad. Nothing currently reports them.

src/DotNetProjectFile.RuleCatalog/GlobalConfigurationEntry.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// <inheritdoc />
+    [Pure]
+    public override string ToString()
+    {
+        var sb = new StringBuilder();
+
+        var param = $"dotnet_diagnostic.{Id}.severity";
+
+        sb.Append($"{param,-35} = {Severity.ToString().ToLowerInvariant(),-10}");
```
JohnVerheij, line 60: The `,-10` pads the severity, but a comment is only appended when `Title` or `Justification` is set. `WRN("IDE0001")` creates an entry with a null `Title`, since that id is not in the collection, so the padding is left dangling and the line ends in spaces. Trimming the end when no comment follows fixes it.

Separately, `Title` is interpolated straight into the line here. Nothing in the current data contains a newline, but a title with one would split a config entry across lines.

specs/Specs/Diagnostics/AnalyzersInfo_specs.cs
```diff
@@ -0,0 +1,119 @@
...
+namespace Diagnostics.AnalyzersInfo_specs;
+
+public class Embedded
+{
+    [Test]
+    public void Contains_rules()
+    {
+        var info = DiagnosticCollection.Embedded();
+        info.Packages.Should().AllSatisfy(p => p.Rules.Should().NotBeEmpty(because: p.Id));
+        info.Count.Should().Be(7437);
```
JohnVerheij: This has to be hand-updated whenever `Collects.New_rules` runs, which is the documented way to refresh the data. A lower bound would keep the intent without the churn.

Concretely: the corpus has `DotNetProjectFile.Analyzers` at 1.17.0 with 180 rules, which is right, since that is the latest release. But `Proj0055` and `Proj0056` from #948 and `Proj6010` from #950 are already on `main`, so the next release makes a refresh necessary, and that is the moment this line and the package list below need editing.

specs/Specs/Diagnostics/AnalyzersInfo_specs.cs
```diff
@@ -0,0 +1,119 @@
...
+    public void Contains_packages()
+    {
+        var info = DiagnosticCollection.Embedded();
+        var packages = info.Packages.Select(p => p.Id);
+
+#if DEBUG
+        Console.WriteLine(string.Join(",\n", packages.Select(p => $"\"{p}\"")));
+#endif
+
+        packages.Should().BeEquivalentTo(
```
JohnVerheij: Same maintenance point as the count: a 74-entry literal to regenerate on every refresh. The `#if DEBUG` `Console.WriteLine` above, which prints the list in exactly the right shape, is the tell. Asserting that every package has rules would keep the intent.

specs/Specs/Diagnostics/AnalyzersInfo_specs.cs
```diff
@@ -0,0 +1,119 @@
...
+[Explicit("Long running process that alters the embedded resource")]
+public class Collects
+{
+    [Test]
+    public async Task New_rules()
+    {
+        var info = DiagnosticCollection.Embedded();
+        info = await DiagnosticCollector.Collect(info);
+
+        var file = new DirectoryInfo("../../../../../src/DotNetProjectFile.Diagnostics/Data/DiagnosticCollection.json");
```
JohnVerheij: Two small things: this is a file path held in a `DirectoryInfo`, and the `../../../../../` is tied to `bin/Release/net10.0`, so it moves if the output path or TFM ever changes.

specs/Specs/GlobalConfig_specs.cs
```diff
@@ -0,0 +1,150 @@
...
+        NON("SA1637"),
+        NON("SA1640"),
+        NON("SA1649"),
+
+        NON("SA1401"),
+        NON("SA1407"),
+        NON("SA1512"),
+        NON("SA1515"));
+
+        config.Save(new FileInfo("../../../../../.globalconfig"), false);
```
JohnVerheij, line 148: Same relative-path anchoring as in `AnalyzersInfo_specs`. Worth pinning both to something stable, since this one writes the repo's `.globalconfig`.

src/DotNetProjectFile.Diagnostics/SimarRules.cs
```diff
@@ -0,0 +1,13 @@
+using System.Collections.Immutable;
+using System.Text.Json.Serialization;
+
+namespace DotNetProjectFile.Diagnostics;
+
+public sealed record SimarRules
```
JohnVerheij: Reads as a typo for `SimilarRules`, type and file name. You mentioned some public names may still change, so flagging it before the package ships.

src/DotNetProjectFile.RuleCatalog/IO/TemporaryDirectory.cs
```diff
@@ -0,0 +1,51 @@
...
+/// Should always be used with a using statement.
+///
+/// <code>
+/// using(var directory = new TemporaryDirectory())
+/// {
+///     // Do stuff.
+/// }
+/// </code>
+/// </remarks>
+internal sealed class TemporaryDirectory : IDisposable
```
JohnVerheij, line 16: File is `TempDirectory.cs`, type is `TemporaryDirectory`.

src/DotNetProjectFile.RuleCatalog/Json/NuGetVersionJsonConverter.cs
```diff
@@ -0,0 +1,39 @@
+using NuGet.Versioning;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace DotNetProjectFile.Diagnostics.Json;
+
+internal sealed class NuGetVersionJsonConverter : JsonConverter<NuGetVersion?>
```
JohnVerheij, line 7: File is `NuGetVersionConverter.cs`, type is `NuGetVersionJsonConverter`.

src/DotNetProjectFile.RuleCatalog/DiagnosticCollector.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </returns>
+    [Pure]
+    public static async Task<DiagnosticCollection> Collect(DiagnosticCollection collection, CancellationToken cancellation = default)
+    {
+        var packages = new Dictionary<string, NugetPackage>();
+
+        foreach (var package in collection.Packages)
+        {
+            var rules = packages.TryGetValue(package.Id, out var existing)
+                ? existing.Rules.ToDictionary(r => r.Id, r => r)
```
JohnVerheij, line 28: `packages` is only written at the end of each iteration, so this lookup can only succeed when `collection.Packages` holds two entries with the same id; otherwise the `existing` branch is unreachable.

The `ToDictionary(r => r.Id, r => r)` on the next line also throws if a package ever ships two descriptors with one id. None of the 74 currently do, so that one is latent rather than live.

src/DotNetProjectFile.RuleCatalog/NuGetRepository.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        await resource.CopyNupkgToStreamAsync(package.Id, package.Version, stream, context, NullLogger.Instance, cancellation);
+
+        using var packageReader = new PackageArchiveReader(stream);
+
+        using var dir = new TemporaryDirectory();
+
+        foreach (var file in packageReader.GetFiles())
+        {
+            packageReader.ExtractFile(file, Path.Combine(dir.FullName, file), NullLogger.Instance);
```
JohnVerheij, line 73: The target path is built from entry names inside a downloaded package. Worth confirming path escapes are rejected somewhere, given the input is arbitrary packages from nuget.org.

src/DotNetProjectFile.RuleCatalog/NuGetRepository.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        static DiagnosticAnalyzer? Analyzers(Type type)
+        {
+            if (!IsDiagnosticAnalyzer(type)) return null;
+
+            try
+            {
+                return Activator.CreateInstance(type) as DiagnosticAnalyzer;
+            }
+            catch
```
JohnVerheij, line 163: The blanket `catch` also swallows `OperationCanceledException`, so cancellation will not surface from these paths. The same applies to the catch in `SupportedDiagnostics` just below.

src/DotNetProjectFile.Diagnostics/IO/TempDirectory.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    [DebuggerBrowsable(DebuggerBrowsableState.Never)]
+    private bool IsDisposed;
+
+    /// <summary>Disposes the temporary directory by deleting it and its content.</summary>
+    public void Dispose()
+    {
+        if (!IsDisposed)
+        {
+            Root.Delete(true);
```
JohnVerheij: `Root.Delete(true)` is unguarded, so a held handle during cleanup throws out of the `using` and masks whatever was in flight.

src/DotNetProjectFile.Diagnostics/NuGetRepository.cs
```diff
@@ -0,0 +1,181 @@
...
+        foreach (var dll in dlls)
+        {
+            using var stream = new MemoryStream();
+            using var reader = dll.OpenRead();
+
+            await reader.CopyToAsync(stream, cancellation);
+
+            try
+            {
+                assemblies.Add(Assembly.Load(stream.ToArray()));
```
JohnVerheij: This loads every downloaded analyzer into the default load context, which cannot be unloaded, and `Activator.CreateInstance` further down runs their constructors. A collectible `AssemblyLoadContext` would scope the third-party code and let the assemblies go once a package is done.

JohnVerheij (commented):
Had a proper look at the branch: built it locally (SDK 10.0.400, Release), ran the spec suite, and poked at the embedded data. Everything below is checked rather than eyeballed, and where I could not confirm something I have said so.

Two things up front so they do not muddy the rest: the three failing specs (`on_global_package_with_custom_license`, `on_package_with_custom_license_in_props`, `transient_dependencies_with_custom_license`) fail on `main` as well, so they are not yours. And comparing a full rebuild of `main` against the branch, the only new warning category is `NU1504`; nothing else is silenced or added. Checked against `main` as of 8c5e685, and the branch still merges cleanly.

I like the direction a lot. Most of the line notes are about the data and the generator rather than the idea. Two things are about the whole corpus rather than any one line.

**41% of the corpus cannot be configured, and that explains the missing titles.** 3,054 of 7,437 rules carry the `NotConfigurable` custom tag, which Roslyn uses precisely to mean the severity cannot be set from an analyzer config. They are almost all compiler diagnostics: 1,611 from `Microsoft.CodeAnalysis.CSharp` and 1,358 from `Microsoft.CodeAnalysis.VisualBasic`, 97% between them. The correlation with empty titles is exact:

```
missing title                : 2971
of those NotConfigurable     : 2971
missing title, configurable  :    0
```

So nothing is being lost on rules that matter. Their titles are empty because the compiler resolves them from resources the reflection-based collector cannot reach, and they are entries a `.globalconfig` generator can never act on anyway. Filtering `WellKnownDiagnosticTags.NotConfigurable` when building configuration entries would remove the dead weight and the empty-title puzzle in one move. For the searchable page they may still be worth keeping, but flagged.

**Three questions about the shape.**

Which of the six capabilities run inside the analyzer? The project is `net10.0` and pulls in `NuGet.Protocol` and `NuGet.Packaging`, and today only `Specs` references it. Items 1-4 read like analyzer rules, which means loading as `netstandard2.0` without the NuGet stack in the compiler process. Splitting a dependency-free data layer (model plus embedded JSON, `netstandard2.0`) from the collector (`net10.0`, NuGet plus reflection) would let 1-4 use the former and keep 5-6 on the latter. Is that the plan, or is the data meant to reach the analyzer another way?

`Contradict` has no description, `Similar` does. A contradiction needs the reason more than a similarity does, since the fix is "pick one" and the user has to know on what grounds.

How does the data stay fresh? `Collect` reaches out to nuget.org, so the file is a point-in-time snapshot and items 1-3 decay quietly as it ages. Scheduled job, or manual?

**One capability worth considering.** Item 3 lines up with something I hit recently. Across the repositories I maintain, `MeziantouAnalysisMode=all-warnings` with `TreatWarningsAsErrors` had quietly enabled both `MA0214` and `MA0215`, which Meziantou's own docs present as opposite approaches with the advice to enable one or the other. `MA0174` and `MA0175` are the same shape. Of roughly 1,435 findings I had to work through, 1,288 were a single rule, `MA0219`, which ships at Hidden severity; almost all the rest were other rules their authors ship disabled and that only appeared because of the blanket switch.

`DiagnosticInfo` already captures `IsEnabledByDefault` and `DefaultSeverity`, so the data to detect that is there. A rule along the lines of "an analysis-mode switch has enabled rules their authors ship disabled" would compose with items 3 and 4. If the `contradict` list is useful to seed, MA0214/MA0215 and MA0174/MA0175 are two documented pairs.

Corniel:
> Two things up front so they do not muddy the rest: the three failing specs (`on_global_package_with_custom_license`, `on_package_with_custom_license_in_props`, `transient_dependencies_with_custom_license`) fail on `main` as well, so they are not yours. And comparing a full rebuild of `main` against the branch, the only new warning category is `NU1504`; nothing else is silenced or added. Checked against `main` as of [8c5e685](https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/commit/8c5e6857246573b28f79da63a41364621bdc5c79), and the branch still merges cleanly.

They fail on my machine too. I still do not fully understand why these extra transient dependencies act differently on the build server and locally, but we should ignore them here indeed.

> **41% of the corpus cannot be configured, and that explains the missing titles.** 3,054 of 7,437 rules carry the `NotConfigurable` custom tag, which Roslyn uses precisely to mean the severity cannot be set from an analyzer config. They are almost all compiler diagnostics: 1,611 from `Microsoft.CodeAnalysis.CSharp` and 1,358 from `Microsoft.CodeAnalysis.VisualBasic`, 97% between them. The correlation with empty titles is exact:
> 
> ```
> missing title                : 2971
> of those NotConfigurable     : 2971
> missing title, configurable  :    0
> ```
> 
> So nothing is being lost on rules that matter. Their titles are empty because the compiler resolves them from resources the reflection-based collector cannot reach, and they are entries a `.globalconfig` generator can never act on anyway. Filtering `WellKnownDiagnosticTags.NotConfigurable` when building configuration entries would remove the dead weight and the empty-title puzzle in one move. For the searchable page they may still be worth keeping, but flagged.

See #955 Thanks for pointing this out. This deserves another rule. As it not obvious when setting it up, so being warned that your configuration is ignored might help.

> Which of the six capabilities run inside the analyzer? The project is `net10.0` and pulls in `NuGet.Protocol` and `NuGet.Packaging`, and today only `Specs` references it. Items 1-4 read like analyzer rules, which means loading as `netstandard2.0` without the NuGet stack in the compiler process. Splitting a dependency-free data layer (model plus embedded JSON, `netstandard2.0`) from the collector (`net10.0`, NuGet plus reflection) would let 1-4 use the former and keep 5-6 on the latter. Is that the plan, or is the data meant to reach the analyzer another way?

That was not final yet, But I was thinking about a way to extra that data that is required for specif rules to create so that they can operate on their own. But the generator should be shipped (one day) as separate package, and be able to fully rely on a modern .NET version.

> `Contradict` has no description, `Similar` does. A contradiction needs the reason more than a similarity does, since the fix is "pick one" and the user has to know on what grounds.

Those analysis are done by asking an LLM to do first suggestion. And `Contradict` still needs that update to have a title/description too.

> How does the data stay fresh? `Collect` reaches out to nuget.org, so the file is a point-in-time snapshot and items 1-3 decay quietly as it ages. Scheduled job, or manual?

I think now and than do a re-scan. We can not rely on LLMs only here, we have to think for our self if they really do the same, or contradict and if we have a preference that we want to advise (for rules that are similar at least)

This PR is to some extend still in the explore phase. It opens up new opportunities. Key here is to have a first version (where the package/project name is the hardest thing to change in the future) might be the most important thing to start with. Other things can be changed while working on it.

src/DotNetProjectFile.RuleCatalog/DiagnosticId.cs
Corniel: You're right. Has been fixed.

src/DotNetProjectFile.Diagnostics/DiagnosticInfo.cs
Corniel: Solved

specs/Specs/Specs.csproj
Corniel: A rebase issue I assume. Thanks for spotting it.

.globalconfig
Corniel: We have to come with a better way of dealing with this. Im not sure what the best approach will be.

specs/Specs/Diagnostics/AnalyzersInfo_specs.cs
Corniel: I've thought about this too. Eventually I want to add our own rules differently than the others, as we have the source on it. But for now I did not take the time yet to do so.

specs/Specs/Diagnostics/AnalyzersInfo_specs.cs
Corniel: Only when the depth of the output path changes. That has not been the case for a long time. So for now, I do not see that as a risk and other approaches I could think of are even less flexible. The test is also only run manually, so it will not unintentionally break things.

src/DotNetProjectFile.Diagnostics/SimarRules.cs
Corniel: This is indeed a typo. Welcome to the world of a dyslectic developer. :)

src/DotNetProjectFile.RuleCatalog/DiagnosticId.cs
Corniel: Solved by the other sorting solves

Issue #942, opened 2026-08-26
Advise on the use of global.json


Corniel:
There should be something in like this:

``` json
{
  "sdk": {
    "version": "8.0.400",
    "rollForward": "latestPatch"
  }
}
```

See: https://learn.microsoft.com/en-us/dotnet/core/tools/global-json

PR #943, opened 2026-08-26
Added a JSON parser

Corniel:
To be able to implement rules about `global.json` (See #942 ) we need a JSON model and parser similar to what we've done with INI files.


specs/Specs/JSON/JsonFile_specs.cs
```diff
@@ -0,0 +1,97 @@
+using DotNetProjectFile.Json;
+
+namespace JSON.JsonFile_specs;
+
+public class Parses
```
CptWesley, line 6: Perhaps add some more cases with edge cases (escaped quotes, newline characters, etc.)

CptWesley (approved):

specs/Specs/JSON/JsonFile_specs.cs
```diff
@@ -0,0 +1,97 @@
...
+                "arr": [1, "two"],
+                "obj": {}
+            }
+            """);
+        var json = JsonFile.Parse(tree);
+
+        var obj = (JsonObject)json.Value!;
+        obj.Properties.Should().HaveCount(7);
+
+        var propStr = obj.Properties.ElementAt(0);
```
Sjaaky: I don't think the order should be something to rely on. Then again, easy change once the implementation changes.

Sjaaky: Also, can't this be expressed like 
```
obj.Properties.Should().BeEquivalent([
   {Key: "str", Text: "hello"},
   {Key: "num", Text: "-12.34e5"},
   ..
  ])
```
Although we miss the typechecking in that case.

specs/Specs/JSON/JsonFile_specs.cs
Corniel: Done

PR #948, opened 2026-08-28
Enable nullability analysis

Corniel:
It turned out that we missed a rule that ensures that nullability analysis is enabled. We sincerely apologize for that. We separated the rule in two, as F# has less flavors and therefor a less smooth migration path.


src/DotNetProjectFile.Analyzers/MsBuild/Nullable.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public enum Kind
+    {
+        /// <summary>Nullability analysis disabled.</summary>
+        Disabled = 0,
+
+        /// <summary>Nullability analysis enabled.</summary>
+        Enabled,
+
+        /// <summary>Nullability analysis enabled.</summary>
```
LauraXiulan: Same as Enabled?

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/EnableNullabilityCSharp.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    /// <inheritdoc />
+    public override ImmutableArray<AnalyzerType> ApplicableTo => ProjectFileTypes.ProjectFile;
+
+    /// <inheritdoc />
+    protected override void Register(ProjectFileAnalysisContext context)
+    {
+        var prop = context.File.Property<Nullable>();
+
+        if (prop is null || prop.Value is not Nullable.Kind.Enabled and not Nullable.Kind.Annotations)
```
LauraXiulan: Why the Annotations check here and not in F#?

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
...
+### 1. Adopt a *Pay-As-You-Go* Strategy (File-by-File)
+Enable analysis in the specific files you are actively working by setting
+`#nullable enable` at the very top of a file when you are fixing a bug or
+adding a feature to it. Only fix the warnings within that single file.
+
+### 2. Prioritize Your Public API and Data Models
+If the project is a shared library or has a core data layer (like EF Core
+models), migrate those files first.Annotating your data models immediately
+clarifies what the database expects.Annotating public methods protects the
+boundaries of your old system from new code.
```
LauraXiulan:
```suggestion
models), migrate those files first. Annotating your data models immediately
clarifies what the database expects. Annotating public methods protects the
boundaries of your old system from new code.
```

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
...
+When migrating old code, if you see a warning but the code has run fine in
+production for five years, do not rewrite the logic. Simply apply the `!`
+operator to tell the compiler: "This works, leave it alone."
+
+```csharp
+// Old code that is proven to work
+string value = GetLegacyData()!;
+```
+
+In all scenario’s you set the global project setting to `<Nullable>annotations</Nullable>`.
```
LauraXiulan: If you do a gradual migration, or in all cases? Not clear from the sentence.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/EnableNullabilityCSharp.cs
CptWesley: F# only supports enabled/disabled

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
+---
```
CptWesley, line 1: What about `annotations` and `warnings`?

src/DotNetProjectFile.Analyzers/MsBuild/Nullable.cs
Corniel:
```suggestion
        /// <summary>Nullability analysis enabled without warning on assignements.</summary>
```

docs/rules/Proj0055.md
Corniel: in all cases. Do you have a suggestion how that can be made clearer?

docs/rules/Proj0055.md
Corniel: It is written in such a way that it should explain that `annotations` is okay for migration scenarios. But do you disagree with that, or do you want to phrase it differently?

docs/rules/Proj0055.md
CptWesley: If it's a new project or everything is properly annotated, we should promote setting it to `enable` rather than `annotations`. This phrashing suggests to always use `annotations`.

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
...
+When migrating old code, if you see a warning but the code has run fine in
+production for five years, do not rewrite the logic. Simply apply the `!`
+operator to tell the compiler: "This works, leave it alone."
+
+```csharp
+// Old code that is proven to work
+string value = GetLegacyData()!;
+```
+
+In all scenario’s you set the global project setting to `<Nullable>annotations</Nullable>`.
```
CptWesley: Maybe this deserves it's own header and more explanation

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+## Compliant
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <Nullable>enable</Nullable>
```
CptWesley, line 53: show example that annotations is also compliant, refer to earlier section about why it's maybe not preferential

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+Or:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <Nullable>disable</Nullable>
```
CptWesley, line 41: show that `warnings` is also not compliant

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
+---
+parent: General
+ancestor: MSBuild
+permalink: /rules/Proj0055
+---
+
+# Proj0055: Enable nullability analysis for C#
+In C# 8.0 and later, [nullability analysis](https://learn.microsoft.com/dotnet/csharp/language-reference/builtin-types/nullable-reference-types)
```
CptWesley, line 8: perhaps also explain somewhere why `warnings` is not desirable (potentially incorrect assumptions about nullability of parameters, no support for `?` and `!`, etc.)

docs/rules/Proj0055.md
```diff
@@ -0,0 +1,81 @@
...
+provides compile-time checks to ensure reference types are safely handled,
```
Sjaaky, line 9: Maybe also point to this documentation about migration:
https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/update-applications/nullable-migration-strategies

src/DotNetProjectFile.Analyzers/MsBuild/Nullable.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public enum Kind
+    {
+        /// <summary>Nullability analysis disabled.</summary>
+        Disabled = 0,
+
+        /// <summary>Nullability analysis enabled.</summary>
+        Enabled,
+
+        /// <summary>Nullability analysis enabled without warning on assignements.</summary>
```
Sjaaky:
```suggestion
        /// <summary>Nullability analysis enabled without warning on assignments.</summary>
```

docs/rules/Proj0055.md
Sjaaky: MS is more forgiving and provides 2 ways for migration, `annotations` and `warnings`.
I'm not sure if one of those is more preferable. That's why I think this rule should be more strict and should only allow 'enabled'.

docs/rules/Proj0055.md
Corniel: I've updated the documentation and just refer to this documentation.

docs/rules/Proj0055.md
Corniel: I've decided that is MS thinks it is a good strategy, I'm not gonna fight it.

docs/rules/Proj0055.md
Corniel: done.

docs/rules/Proj0055.md
Corniel: I decided to only disallow `disabled`.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/EnableNullabilityFSharp.cs
```diff
@@ -0,0 +1,24 @@
+using Nullable = DotNetProjectFile.MsBuild.Nullable;
+
+namespace DotNetProjectFile.Analyzers.MsBuild;
+
+/// <summary>Implements <see cref="Rule.EnableNullabilityFSharp" />.</summary>
+[DiagnosticAnalyzer(LanguageNames.CSharp, LanguageNames.VisualBasic)]
```
Sjaaky, line 6: Shouldn't this be 
```suggestion
[DiagnosticAnalyzer(LanguageNames.FSharp)]
```
?

Corniel: No. FSharp does is not compiled using Roslyn. Technically, VB can be dropped here.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/EnableNullabilityFSharp.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public override ImmutableArray<Language> ApplicableLanguages => Languages.FSharp;
+
+    /// <inheritdoc />
+    public override ImmutableArray<AnalyzerType> ApplicableTo => ProjectFileTypes.ProjectFile;
+
+    /// <inheritdoc />
+    protected override void Register(ProjectFileAnalysisContext context)
+    {
+        var prop = context.File.Property<Nullable>();
```
Sjaaky, line 17: Should there be a check if  F# >= 9.0 is used?

docs/rules/Proj0056.md
```diff
@@ -0,0 +1,50 @@
+---
+parent: General
+ancestor: MSBuild
+permalink: /rules/Proj0056
+---
+
+# Proj0056: Enable nullability analysis for F#
+F# 9.0 introduced support for [nullable reference types](https://learn.microsoft.com/dotnet/fsharp/language-reference/values/null-values)
+to match [C#'s null-safety checks](Proj0055.md) when working with nullable
+types or interacting with C# code.
```
Sjaaky:
```suggestion
to match [C#'s null-safety checks](Proj0055.md) for interoperability with C# code.
```
Nullable types are introduced just for interoperability. 
"Using | null in F# type modeling is not considered idiomatic" (https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/values/null-values)

Sjaaky (approved):
Some small remarks.

src/DotNetProjectFile.Analyzers/Analyzers/MsBuild/EnableNullabilityFSharp.cs
Corniel: That would be a small improvement. However, checking for it is not trivial. Furthermore, as far as I know, there are hardly any issues in updating to F# 10 anyway. So until some reports this with a FP and state that they can not upgrade to F# 10, I'd like to leave it as it is.

PR #950, opened 2026-08-28
Proj6010: global.json should exist

Corniel:
To ensure a predictable .NET SDK version is used across different machines, CI/CD pipelines, and local development environments, a [`global.json`](https://learn.microsoft.com/dotnet/core/tools/global-json) file should exist for every MSBuild project.

When a `global.json` file is missing, the .NET CLI and MSBuild will resolve to the latest installed .NET SDK version on the compiling host machine. This can introduce subtle bugs, build failures, or differences in compiler diagnostics if developers have different SDK versions installed.

## Non-compliant
No `global.json` file is present in the project directory or any parent directories up to the repository root.

## Compliant
A `global.json` file is defined at the project directory level or in a parent directory, specifying the .NET SDK version:

```json
{
  "sdk": {
    "version": "10.0.400",
    "rollForward": "latestPatch"
  }
}
```

See: #942

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -267,6 +267,10 @@ reported to the [GitHub repository](https://github.com/dotnet-project-file-analy
 * [**Proj5005** Omit Project ID's](rules/Proj5005.md)
 * [**Proj5006** Included files should exist](rules/Proj5006.md)
 
+## JSON
+* [**Proj6000** Invalid JSON file](rules/Proj6000.md)
+* [**Proj6010** global.json should exist](rules/Proj6010.md)
+
 ## SonarQube integration
 By default, results by .NET project file analyzers are included in SonarQube
 reports. Read [here](general/sonar-integration.md) more about how this works.
--- a/docs/rules/Proj6000.md
+++ b/docs/rules/Proj6000.md
@@ -1,0 +1,9 @@
+---
+parent: JSON
+ancestor: Rules
+permalink: /rules/Proj6000
+---
+
+# Proj6000: Invalid JSON file
+A [JSON file](https://www.json.org/) is a plain text file with a lightweight
+data-interchange format.
--- a/docs/rules/Proj6010.md
+++ b/docs/rules/Proj6010.md
@@ -1,0 +1,32 @@
+---
+parent: JSON
+ancestor: Rules
+permalink: /rules/Proj6010
+---
+
+# Proj6010: global.json should exist
+To ensure a predictable .NET SDK version is used across different machines,
+CI/CD pipelines, and local development environments, a [`global.json`](https://learn.microsoft.com/dotnet/core/tools/global-json)
+file should exist for every MSBuild project.
+
+When a `global.json` file is missing, the .NET CLI and MSBuild will resolve to
+the latest installed .NET SDK version on the compiling host machine. This can
+introduce subtle bugs, build failures, or differences in compiler diagnostics
+if developers have different SDK versions installed.
+
+## Non-compliant
+No `global.json` file is present in the project directory or any parent
+directories up to the repository root.
+
+## Compliant
+A `global.json` file is defined at the project directory level or in a parent
+directory, specifying the .NET SDK version:
+
+```json
+{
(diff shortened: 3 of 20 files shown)


docs/README.md
```diff
@@ -267,6 +267,10 @@ reported to the [GitHub repository](https://github.com/dotnet-project-file-analy
 * [**Proj5005** Omit Project ID's](rules/Proj5005.md)
 * [**Proj5006** Included files should exist](rules/Proj5006.md)
 
+## JSON
+* [**Proj6000** Invalid JSON file](rules/Proj6000.md)
+* [**Proj6010** global.json should exist](rules/Proj6010.md)
```
Silvex78, line 272: Why not Proj6001?

docs/rules/Proj6000.md
```diff
@@ -0,0 +1,9 @@
+---
+parent: JSON
+ancestor: Rules
+permalink: /rules/Proj6000
+---
+
+# Proj6000: Invalid JSON file
+A [JSON file](https://www.json.org/) is a plain text file with a lightweight
```
Silvex78, line 8: Does this check all JSON files?

specs/Specs/Rules/SLNX/Use_SLNX_files.cs
```diff
@@ -5,15 +5,14 @@ namespace Rules.SLNX.Use_SLNX_files;
...
-    public void missing_SLNX() => new UseSlnxFiles()
-       .ForInlineSdkProject("""
-<Project Sdk="Microsoft.NET.Sdk">
+    public void missing_SLNX() => new UseSlnxFiles().ForInlineSdkProject("""
+        <Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>net10.0</TargetFramework>
-  </PropertyGroup>
+          <PropertyGroup>
```
Silvex78, line 11: Did this file contain spaces/tabs issues?

docs/rules/Proj6010.md
```diff
@@ -0,0 +1,32 @@
...
+directories up to the repository root.
+
+## Compliant
+A `global.json` file is defined at the project directory level or in a parent
+directory, specifying the .NET SDK version:
+
+```json
+{
+  "sdk": {
+    "version": "10.0.400",
```
Silvex78, line 28: How are we going to handle SDK updates? This must now be updated by hand in this file?

docs/rules/Proj6010.md
```diff
@@ -0,0 +1,32 @@
...
+CI/CD pipelines, and local development environments, a [`global.json`](https://learn.microsoft.com/dotnet/core/tools/global-json)
+file should exist for every MSBuild project.
+
+When a `global.json` file is missing, the .NET CLI and MSBuild will resolve to
+the latest installed .NET SDK version on the compiling host machine. This can
+introduce subtle bugs, build failures, or differences in compiler diagnostics
+if developers have different SDK versions installed.
+
+## Non-compliant
+No `global.json` file is present in the project directory or any parent
```
Silvex78, line 18: Do we need to do a suggestion here that a global.json in a mono repo's root might not be the best idea when projects can contain multiple SDK version?

Silvex78 (approved):

docs/rules/Proj6000.md
Corniel: Nope, only those who are added as additional file (which is only `global.json` at the moment)

specs/Specs/Rules/SLNX/Use_SLNX_files.cs
Corniel: just a bit reformatting.

docs/README.md
Corniel: To allow to extend the JSON parsing errors.

docs/rules/Proj6010.md
Corniel: I would argue that that should be up to the development team. We can not draw any conclusion from it.

PR #956, opened 2026-09-02
Proj1008: Remove ineffective rule configurations

### LINKED ISSUE #955: Do not configure NotConfigurable Diagnostics in globalconfig (or editorconfig)
Thanks to @JohnVerheij for pointing this out in #900

Corniel:
I've decided to have one rule for both the MSBuild files, globalconfig files, as C# code (first C# rule to be added). My reasoning for doing so is that the advise is always the same: remove the configuration, and I do not see why you want to suppress this other than a FP (bug) that we should fix.

---------------------
# Proj1008: Remove ineffective rule configurations
Some diagnostic rules are not configurable. Their severity cannot be modified or suppressed. Code that attempts to configure these rules should therefore be removed.

## Non-compliant
Non-compliant code should simply be removed:

### MSBuild files
```xml

  <PropertyGroup>
    <NoWarn>CS0016</NoWarn>
  </PropertyGroup>

</Project>
```

### .globalconfig and .editorconfig files
``` ini
dotnet_diagnostic.CS0016.severity = none
```

### C# Source
```csharp
#pragma warning disable CS0016 // Non-compliant

[SuppressMessage("Category", "CS1680")] // Non-compliant
public class SomeClass
{
}
```

Closes #955


specs/Specs/Specs.csproj
```diff
@@ -52,13 +52,18 @@
     <PackageReference PrivateAssets="all" Include="NUnit3TestAdapter" Version="6.2.0" />
   </ItemGroup>
 
+  <ItemGroup Label="Test case C# files">
+    <!--Compile Remove="Cases/*.*" /-->
```
pmdevers: This can be removed

pmdevers (approved):

PR #958, opened 2026-09-02
Proj1009: Remove configuration for dropped rule

### LINKED ISSUE #957: Remove configuration on dropped rules
Rules that used to exist, but are not longer available in (the latest version of) the analyzers, should not be configured. In some edge cases this could lead to FP's because some projects might still rely on older versions of the analyzer reported on, but I would argue those versions should be updated anyway.

Corniel:
Some diagnostic rules are no longer supported or have been removed by the analyzer creators. Code that attempts to configure these rules should therefore be removed.

## Non-compliant
Non-compliant code should simply be removed:

### MSBuild files
```xml

  <PropertyGroup>
    <NoWarn>RS1038</NoWarn>
  </PropertyGroup>

</Project>
```

### .globalconfig and .editorconfig files
``` ini
dotnet_diagnostic.RS1038.severity = warning
```

### C# Source
```csharp
#pragma warning disable RS1038 // Non-compliant

[SuppressMessage("Category", "RS1038")] // Non-compliant
public class SomeClass
{
}
```
Closes #957


specs/Specs/Cases/Remove_ineffective_rule_configuration.cs
```diff
@@ -3,6 +3,8 @@
 #pragma warning disable S101
 #pragma warning disable CS1680 // Noncompliant {{Rule CS1680 is not-configurable and cannot be modified}}
 //                      ^^^^^^
+#pragma warning disable AK1001 // Noncompliant {{Rule AK1001 is does not longer exist}}
```
Silvex78:
```suggestion
#pragma warning disable AK1001 // Noncompliant {{Rule AK1001 does no longer exist}}
```

specs/Specs/Rules/Remove_ineffective_rule_configuration.cs
```diff
@@ -10,16 +10,17 @@ public class Reports_not_configurable_rule_IDs_in
...
         </Project>
         """)
         .HasIssues(
             Issue.WRN("Proj1008", "Rule CS0016 is not-configurable and cannot be modified").WithSpan(04, 04, 04, 27),
             Issue.WRN("Proj1008", "Rule SA1603 is not-configurable and cannot be modified").WithSpan(05, 04, 05, 52),
-            Issue.WRN("Proj1008", "Rule CS1680 is not-configurable and cannot be modified").WithSpan(06, 04, 06, 67),
-            Issue.WRN("Proj1008", "Rule CS1681 is not-configurable and cannot be modified").WithSpan(06, 04, 06, 67));
+            Issue.WRN("Proj1008", "Rule CS1680 is not-configurable and cannot be modified").WithSpan(06, 04, 06, 74),
+            Issue.WRN("Proj1008", "Rule CS1681 is not-configurable and cannot be modified").WithSpan(06, 04, 06, 74),
+            Issue.WRN("Proj1009", "Rule AK1001 is does not longer exist").WithSpan(06, 04, 06, 74));
```
Silvex78:
```suggestion
            Issue.WRN("Proj1009", "Rule AK1001 does no longer exist").WithSpan(06, 04, 06, 74));
```

specs/Specs/GlobalConfig_specs.cs
```diff
@@ -143,7 +143,8 @@ public void Solution_wide_preferences()
         NON("SA1401"),
         NON("SA1407"),
         NON("SA1512"),
-        NON("SA1515"));
+        NON("SA1515"),
+        NON("RS1038", "VB.NET specific dependecty is not needed"));
```
Silvex78:
```suggestion
        NON("RS1038", "VB.NET specific dependency is not needed"));
```

src/DotNetProjectFile.Analyzers/Rule.cs
```diff
@@ -1346,12 +1346,22 @@ public static partial class Rule
...
-            "Configuring severity or suppression of non-configurable rules, " +
+            "Configuring severity or suppression of non-configurable rules " +
+            "are redundant and should be removed.",
+        tags: ["suppression", "severity", "warning", "error"],
+        category: Category.Configuration);
+
+    public static DiagnosticDescriptor RemoveDroppedRuleConfiguration => New(
+        id: 1009,
+        title: "Remove configuration for dropped rule",
+        message: "Rule {0} is does not longer exist",
```
Silvex78:
```suggestion
        message: "Rule {0} does no longer exist",
```

src/DotNetProjectFile.RuleCatalog/DiagnosticInfo.cs
```diff
@@ -93,17 +96,29 @@ public DiagnosticInfo Update(DiagnosticInfo update) => update with
...
         Description = Description.NullIfEmpty(),
         HelpLinkUri = HelpLinkUri.NullIfEmpty(),
         Obsolete = Obsolete.NullIfEmpty(),
-    };
+    }).Drop();
+
+    /// <summary>Set a dropped message if applicable.</summary>
+    [Pure]
+    private DiagnosticInfo Drop()
+        => Version is not null && Obsolete is null or "This rule is deprecated."
```
Silvex78, line 111: I have a hard time understanding this condition. With a version and no obsolete the rule has been dropped?

Silvex78 (approved):

src/DotNetProjectFile.RuleCatalog/DiagnosticInfo.cs
Corniel: No. If the current `Version` is set at this stage, it means it is different (lower) that current `Version` of the package. So if no obsolete message is provided (or `This rule is deprecated.`) it is better to update that the rule has been dropped (which is stronger than deprecated).

PR #969, opened 2026-09-09
Loading/collection rules faces issues

Corniel:
- [x] Update definitions based on own code
- [x] Solve assembly and type load issues


CptWesley:
What issue is this exactly tackling?

Corniel:
@CptWesley some dlls are not correctly resolved from a package as depedend other dll's shipped too.

